PR #8, opened 2016-11-07
MOBILE-2292: Add JSON deserialization to JsonSerializable

brianblanchard-wf:
### Description
- To do some of the serialization of w_modules there has to be some kind of common serialization logic that classes can implement. Add a deserialization constructor to the existing JsonSerializable abstract class so all consumers of w_common have access to it.

### Changes
- Added a `fromJson` named constructor to the `JsonSerialization` abstract class.
- Updated unit tests and documentation.

### Semantic Versioning

> **This library is still pre-1.0.0.**
>
> Patches and minor changes will be released in a patch version, while breaking
> changes can be released in a minor version.

- [x] **Patch**
- [ ] **Minor**

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf @evanweible-wf @jayudey-wf @maxwellpeterson-wf @sebastianmalysa-wf @trentgrover-wf


aviary2-wf:
## Raven
Number of Findings: 0

dustinlessard-wf:
+1

evanweible-wf:
@brianblanchard-wf unfortunately I don't think this is having the intended effect - it doesn't force extending/implementing classes to provide a `fromJson()` constructor. Regardless of whether a subclasser `implements` or `extends` JsonSerializable, their class can provide any (or no) constructors. For example:

```dart
abstract class JsonSerializable {
  JsonSerializable();
  JsonSerializable.fromJson(Map json);
}

class ImplementingClass implements JsonSerializable {}

class ExtendingClass extends JsonSerializable {
  ExtendingClass(String differentSignature) { /* ... */ }
  ExtendingClass.fromJson(String encodedJson) { /* ... */ }
}
```

If you were to run the above code through the `dartanalyzer`, it would come back clean. It appears there's no way in Dart to guarantee that a class will provide a certain constructor (this is not something I totally realized until just playing around with it today, so if someone knows something I don't, please correct me).

maxwellpeterson-wf:
Building on what Evan said - I was curious if this was possible in Java (which has heavily influenced Dart) and it seems like it isn't supported there either: http://stackoverflow.com/questions/3164334/how-can-i-force-a-constructor-to-be-defined-in-all-subclass-of-my-abstract-class

This is something I'd be curious to get the Dart language guys' opinions on. Perhaps there is an alternative approach that yields a similar result.

brianblanchard-wf:
@maxwellpeterson-wf I was thats what I was running into as well. I guess I can define a `fromJson` instance method that you could call after initializing a new object? Thoughts?

```
JsonSerializable()..fromJson(map)
```

evanweible-wf:
@brianblanchard-wf I'm not sure I like that approach either. Beyond the fact that it's a method playing the role of a constructor, it would prevent an extending/implementing class from being immutable.

Perhaps there's another approach or solution that would suffice here. Could you require that consumers of your API provide a separate factory function that takes a serialized input?

brianblanchard-wf:
@evanweible-wf Wouldn't that lead to the same issue as the named constructors? I am pretty sure you cannot enforce any type of constructor. You guys are much more proficient in dart than I am so any other suggestions are appreciated.

brianblanchard-wf:
@evanweible-wf I have seen the named constructor pattern used in the dart transformer code. It is very similar to how I am using the named constructor.

Definition: https://github.com/dart-lang/barback/blob/master/example/simple_transformer/lib/transformer.dart#L12-L15

Invocation: https://github.com/dart-lang/pub/blob/3a96c86ab75255e114eb47aaadf4f1f81cf056b7/lib/src/asset/dart/transformer_isolate.dart#L69

evanweible-wf:
@brianblanchard-wf yes, it would be possible with mirrors. My suggestion above may not have been clear, I'm not saying that you require a factory constructor, instead your could require a function matching a certain signature that returns an instance of whatever it is you would want to be the result of the deserialization.

```dart
typedef SerializableEvent EventDeserializer(Map<String, dynamic> json);
```

Then your API would expect something of that type in order to do the deserialization. Does that make sense?

brianblanchard-wf:
After talking with @evanweible-wf offline, it looks like this approach is not going to have the desired effects. Closing this CR.

PR #14, opened 2016-11-14
REF-1399 Fail immediately on null argument

georgelesica-wf:
### Description

Right now we fail if one of the "manage" methods on `Disposable` gets a `null` value. However, we only fail once `dispose()` has been called, so the error is rather confusing.

### Changes

It seems reasonable to me that those methods shouldn't accept `null` arguments because there's no reason for that to ever occur in actual code, and if it did, it's almost certainly a bug in the consumer's code. If we don't deal with the `null` noisily it could lead to tricky memory leaks or other hard-to-find errors.

However, we should fail immediately so that we provide the consumer with an easily-understood stack trace and error message. This PR just makes those methods throw if they get a `null` argument.

### Semantic Versioning

> **This library is still pre-1.0.0.**
>
> Patches and minor changes will be released in a patch version, while breaking
> changes can be released in a minor version.

- [x] **Patch**
- [ ] **Minor**

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf @evanweible-wf @jayudey-wf @maxwellpeterson-wf @sebastianmalysa-wf @trentgrover-wf

### FYI

@dustyholmes-wf

### DIFF
--- a/lib/disposable.dart
+++ b/lib/disposable.dart
@@ -138,23 +138,32 @@ abstract class Disposable implements _Disposable {
   }
 
   /// Automatically dispose another object when this object is disposed.
+  ///
+  /// The parameter may not be `null`.
   @mustCallSuper
   @protected
   void manageDisposable(Disposable disposable) {
+    _throwIfNull(disposable, 'disposable');
     _internalDisposables.add(disposable);
   }
 
   /// Automatically handle arbitrary disposals using a callback.
+  ///
+  /// The parameter may not be `null`.
   @mustCallSuper
   @protected
   void manageDisposer(Disposer disposer) {
+    _throwIfNull(disposer, 'disposer');
     _internalDisposables.add(new _InternalDisposable(disposer));
   }
 
   /// Automatically cancel a stream controller when this object is disposed.
+  ///
+  /// The parameter may not be `null`.
   @mustCallSuper
   @protected
   void manageStreamController(StreamController controller) {
+    _throwIfNull(controller, 'controller');
     _internalDisposables.add(new _InternalDisposable(() {
       if (!controller.hasListener) {
         controller.stream.listen((_) {});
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

maxwellpeterson-wf:
+1

evanweible-wf:
+1

lib/disposable.dart
```diff
@@ -181,4 +193,10 @@ abstract class Disposable implements _Disposable {
     _didDispose.complete();
     return null;
   }
+
+  void _throwIfNull(dynamic subscription, String name) {
```
dustinlessard-wf: Can we consider making this a utility method exposed form this library? I could see uses for this beyond a 'subscription'.

dustinlessard-wf:
+1

lib/disposable.dart
dustyholmes-wf: I think ArgumentError.notNull is the majority of what a utility could be anyway. I haven't run into much use for a re usable error throwing helper.

dustinlessard-wf: I was thinking something like this:
```dart
void doSomething(AwesomeObject obj) {
     // do some expensive things.
     obj.callMethod();
}
```
That could be rewritten to avoid unnecessarily doing anything expensive:
```dart
void doSomething(AwesomeObject obj) {
     throwIfNull(obj, 'doSomething');
     // do some expensive things.
     obj.callMethod();
}
```

georgelesica-wf: If the actual solution was a bit more elegant I would definitely support pulling it into w_common proper. But as it stands, it's not. Hopefully non-nullable types will save us from this nightmare... https://github.com/dart-lang/sdk/issues/27231 :wink:

georgelesica-wf:
@jayudey-wf I think we're ready for merge.

jayudey-wf:
QA Resource Approval: +10

- [ ] Testing instruction
- [ ] Dev +1's
 - passing CI exercising new functionality
- [ ] Unit test created/updated
- [ ] All unit tests pass

Merging into master.

PR #17, opened 2016-12-27
RAP-1500 Extract manage methods into an interface

georgelesica-wf:
### Description

We want to be able to expose the helpful manage methods without making a class itself "Disposable".

### Changes

Extract manage methods into an interface that we then implement in `Disposable`. Additionally, remove `@protected` from the manage method implementations. We need this for two reasons. First, because they're not technically part of the interface if they're protected, and second, because we want to be able to use `Disposable` as a concrete proxy. Finally, make `Disposable` concrete.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf
@bencampbell-wf 
@aaronstgeorge-wf

### DIFF
--- a/lib/disposable.dart
+++ b/lib/disposable.dart
@@ -120,11 +120,15 @@ typedef Future<dynamic> Disposer();
 ///
 /// Below is an example of using the class as a concrete proxy.
 ///
-///      class MyLifecycleThing {
+///      class MyLifecycleThing implements DisposableManager {
 ///        Disposable _disposable = new Disposable();
 ///
 ///        MyLifecycleThing() {
 ///          _disposable.manageStreamSubscription(someStream.listen(() => null));
+///        }
+///
+///        void manageStreamSubscription(StreamSubscription sub) {
+///          _disposable.manageStreamSubscription(sub);
 ///        }
 ///
 ///        // ...
--- a/lib/disposable.dart
+++ b/lib/disposable.dart
@@ -127,11 +127,12 @@ typedef Future<dynamic> Disposer();
 ///          _disposable.manageStreamSubscription(someStream.listen(() => null));
 ///        }
 ///
+///        @override
 ///        void manageStreamSubscription(StreamSubscription sub) {
 ///          _disposable.manageStreamSubscription(sub);
 ///        }
 ///
-///        // ...
+///        // ...more methods
 ///
 ///        Future<Null> unload() async {
 ///          await _disposable.dispose();
--- a/lib/disposable.dart
+++ b/lib/disposable.dart
@@ -37,13 +37,25 @@ class _InternalDisposable implements _Disposable {
   }
 }
 
+/// An interface to abstract management from actual disposal machinery.
+///
+/// When new management methods are to be added, they should be added
+/// here first, then implemented in [Disposable].
+abstract class DisposableManager {
+  void manageDisposable(Disposable disposable);
+  void manageDisposer(Disposer disposer);
+  void manageStreamController(StreamController controller);
+  void manageStreamSubscription(StreamSubscription subscription);
+}
+
 /// A function that, when called, disposes of one or more objects.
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/disposable.dart
```diff
@@ -37,13 +37,25 @@ class _InternalDisposable implements _Disposable {
   }
 }
 
+/// An interface to abstract management from actual disposal machinery.
```
dustinlessard-wf: This read a bit weird to me. If anyone else agrees, can we consider tweaking it?

evanweible-wf:
So if we remove the `@protected` annotations and a consumer uses the `Disposable` class as a mixin, they would have to override and then add the `@protected` annotations themselves in order to prevent external consumers of the extending class from calling the `manage...()` methods, right?

lib/disposable.dart
```diff
@@ -37,13 +37,25 @@ class _InternalDisposable implements _Disposable {
...
 /// A function that, when called, disposes of one or more objects.
 typedef Future<dynamic> Disposer();
 
 /// Allows the creation of managed objects, including helpers for common patterns.
 ///
-/// There are three ways to consume this class: as a mixin, a base class,
-/// and an interface. All should work fine but the first is the simplest
+/// There are four ways to consume this class: as a mixin, a base class,
+/// an interface, and a concrete class used as a proxy. All should work
+/// fine but the first is the simplest
```
dustinlessard-wf: I'm curious why we might see the mixin as the most powerful approach?

georgelesica-wf:
@evanweible-wf correct, although now that those methods are part of an interface, I'm not sure if they should be marked protected since then the class would no longer technically implement the interface. All of this would be much simpler if `@protected` was an actual part of the language 😞 .

lib/disposable.dart
georgelesica-wf: That was already in there, but basically because it requires the least code on the part of the consumer.

lib/disposable.dart
georgelesica-wf: Suggestions?

dustinlessard-wf: A manager for disposable members. The consumer of this interface has precise control over how their disposable members are disposed.

dustinlessard-wf: ^^ meh, could be better but this is my two cents.   Take it or leave it.

dustinlessard-wf:
+1

evanweible-wf:
+1

dustinlessard-wf:
+1

lib/disposable.dart
```diff
@@ -102,7 +117,32 @@ typedef Future<dynamic> Disposer();
...
+/// Below is an example of using the class as a concrete proxy.
+///
+///      class MyLifecycleThing implements DisposableManager {
+///        Disposable _disposable = new Disposable();
+///
+///        MyLifecycleThing() {
+///          _disposable.manageStreamSubscription(someStream.listen(() => null));
+///        }
+///
+///        void manageStreamSubscription(StreamSubscription sub) {
```
aaronstgeorge-wf: This could use an `@override`

georgelesica-wf: Done

dustinlessard-wf:
+1

evanweible-wf:
+1

bencampbell-wf:
+10

Things Tested
-------------
- [x] CI passed.

bencampbell-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [ ] Unit tests created/updated - NA - Added Interface
- [x] All unit tests pass

Merging into master.

PR #19, opened 2017-01-04
RAP-1503 Implement Cache and ReferenceCache

bencampbell-wf:
### Description

The need to remove duplicates from or maintain a collection of instances of a particular class demands the need for a Cache implementation. The Cache should allow consumers to obtain a value from the cache or call an underlying factory, explicitly put an item in the Cache, and remove items from a Cache.

### Changes

Implement the base Cache implementation. The get, getAsync, put, and remove operations allow consumers to interact with the Cache as described. Lifecycle methods and an update stream are provided allowing consumers to extend the cache and implement custom behaviors or observe the cache changes.

A specialization of the Cache is the ReferenceCache which maintains a list of accesses from the Cache. Consumers can release a reference from the cache which decrements the reference count. If the number of releases equals the number of gets, the item is removed from the ReferenceCache. This effectively allows consumers to memoize shared resources.

For example:

```dart

Func<Future<Thing>> buildThingFactory() {
    ReferenceCache<String, HeavyNetworkConnector> cache = new ReferenceCache<String, HeavyNetworkConnector>()
    cache.didRemove.listen((String id, HeavyNetworkConnection connector) {
        print('Last reference to connection $id has been released: disposing');
        connection.dispose();
    });

    return (String thingId) async {
        final connection = await cache.getAsync(thingId, () => _expensiveOpToInitConnection(thingId));
        final thing =  new Thing(networkConnection: connection)
            ..didDispose.then((Null _) => cache.remove(thingId));

        return thing;
    }
}
```

Or you can cache function results via composing a cache:

```dart
class HeavyOp {
    Cache<Token, OpResult> _cache = new Cache<Token, OpResult>();

    Future<OpResult> runOp(Token opParam) async {
        final value = await _cache.get(opToken, () => _executeOp(opParam));
        
        // Cache results are invalidated after thirty seconds.
        new Timer(new Duration(seconds: 30), () {
            _cache.remove(opParam);
        });

        return value
    }

    OpResult _executeOp(Token opParam) {
        /// lengthy computation 
        return result;
    }
}
```

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf
@todbachman-wf 
@georgelesica-wf 
@aaronstgeorge-wf

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * A `Cache` implementation that maintians references to an object instance by
+  * A `Cache` implementation that maintains references to an object instance by
   an identifier. Specializations of this class allow invariants to be
   maintained:
     * `ReferenceCache` maintains a count for each access for a given identifier.
--- a/README.md
+++ b/README.md
@@ -6,14 +6,14 @@ includes the following:
   * A `Cache` implementation that maintains references to an object instance by
   an identifier. Specializations of this class allow invariants to be
   maintained:
-    * `ReferenceCache` maintains a count for each access for a given identifier.
+    * A `ReferenceCache` maintains a count for each access for a given identifier.
     When the final reference is released, the item is removed from the cache.
   * A `Disposable` interface / mixin to assist with cleaning up streams and
   other data structures that won't necessarily be garbage collected without some
   manual intervention.
   * A simple typedef that can be parameterized to represent a zero-arity
   callback that returns a particular type.
-  * A `InvalidationMixin` mixin used to mark a class as requiring validation.
+  * An `InvalidationMixin` mixin used to mark a class as requiring validation.
   * A `JsonSerializable` interface to indicate that something can be serialized
   to JSON.
 
--- a/README.md
+++ b/README.md
@@ -4,10 +4,9 @@ A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
   * A `Cache` implementation that maintains references to an object instance by
-  an identifier. Specializations of this class allow invariants to be
-  maintained:
-    * A `ReferenceCache` maintains a count for each access for a given identifier.
-    When the final reference is released, the item is removed from the cache.
+  an identifier. Specializations of this class allow invariant to be
+  maintained. For instance, the `ReferenceCache` maintains a count of gets
+  and releases. The item is removed from the cache when these counts are equal.
   * A `Disposable` interface / mixin to assist with cleaning up streams and
   other data structures that won't necessarily be garbage collected without some
   manual intervention.
--- a/README.md
+++ b/README.md
@@ -3,13 +3,18 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * `Disposable` interface / mixin to assist with cleaning up streams and other
-  data structures that won't necessarily be garbage collected without some
+  * A `Cache` implementation that maintians references to an object instance by
+  an identifier. Specializations of this class allow invariants to be
+  maintained:
+    * `ReferenceCache` maintains a count for each access for a given identifier.
+    When the final reference is released, the item is removed from the cache.
+  * A `Disposable` interface / mixin to assist with cleaning up streams and
+  other data structures that won't necessarily be garbage collected without some
   manual intervention.
   * A simple typedef that can be parameterized to represent a zero-arity
   callback that returns a particular type.
-  * `InvalidationMixin` mixin used to mark a class as requiring validation.
-  * `JsonSerializable` interface to indicate that something can be serialized
+  * A `InvalidationMixin` mixin used to mark a class as requiring validation.
+  * A `JsonSerializable` interface to indicate that something can be serialized
   to JSON.
 
 We expect this list to grow as we identify small pieces of code that are useful
--- a/lib/src/cache/cache.dart
+++ b/lib/src/cache/cache.dart
@@ -1,0 +1,237 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'package:w_common/disposable.dart';
+import 'package:w_common/func.dart';
+
+/// Immutable payload that indicates a change in a [Cache].
+class CacheContext<TIdentifier, TValue> {
+  /// Identifies a single instance of a [TValue]
+  final TIdentifier id;
+
+  /// The current value stored in the [Cache] for the attributed [TIdentifier].
+  final TValue value;
+
+  CacheContext(this.id, this.value);
+}
+
+/// Maintians a reference to a given any [TValue] attributed to a [TIdentifier].
+///
+/// References are retained for the lifecycle of the instance of the [Cache],
+/// unless explicitly removed.
+class Cache<TIdentifier, TValue> extends Object with Disposable {
+  Map<TIdentifier, Future<TValue>> _cache = <TIdentifier, Future<TValue>>{};
+  StreamController<CacheContext<TIdentifier, TValue>> _didUpdateController =
+      new StreamController<CacheContext<TIdentifier, TValue>>.broadcast();
+  StreamController<CacheContext<TIdentifier, TValue>> _didRemoveController =
+      new StreamController<CacheContext<TIdentifier, TValue>>.broadcast();
+  Map<TIdentifier, Completer<Null>> _removalCompleters =
+      <TIdentifier, Completer<Null>>{};
+
+  Cache() {
+    [_didRemoveController, _didUpdateController]
+        .forEach(manageStreamController);
+  }
+
+  /// The collection of removals from the [Cache].
+  Stream<CacheContext<TIdentifier, TValue>> get didRemove =>
+      _didRemoveController.stream;
+
+  /// The collection of changes to the [Cache].
+  ///
+  /// Values are emitted when the cached value is changes.
+  Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
+      _didUpdateController.stream;
+
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
+  /// [TValue] returned by the value factory when the value does not exist in
+  /// the cache.
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
+  /// returning the cached value. If the [Cache] does not contain an instance
+  /// for the given [TIdentifier], the given [valueFactory] is called and an
+  /// [CacheContext] event is emitted on the [didUpdate] stream. A call to [get]
+  /// that returns a cached value does not emit this event as the [Cache] has
+  /// not updated.
+  ///
+  /// Calls to [get] are evaluated syncronously and will return the future value
+  /// established by the first call to [get]. This allows calls to get to
+  /// return the same value regardless of completion of asyncronous calls to
+  /// the [onGet] lifecycle methods or [valueFactory].
+  ///
+  /// ```
+  /// var a = cache.get('id', _superLongCall);
+  /// var b = cache.get('id', _superLongCall);
+  /// var c = cache.get('id', _superLongCall);
+  ///
+  /// // All of the values in this list are the same instance returned by the
+  /// // first call to `_superLongCall`.
+  /// var values = Future.wait([a, b, c]);
+  /// ```
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  @mustCallSuper
+  Future<TValue> get(TIdentifier id, Func<TValue> valueFactory) {
+    _throwWhenDisposed('get');
+
+    // Await any pending cached futures
+    if (_cache.containsKey(id)) {
+      return _cache[id].then((TValue value) async {
+        await onGet(id, value);
+        return value;
+      });
+    }
+
+    // Install Future value
+    final Completer<TValue> completer = new Completer<TValue>();
+    _cache[id] = completer.future;
+    final TValue value = valueFactory.call();
+
+    onGet(id, value).then((Null _) {
+      _didUpdateController.add(new CacheContext(id, value));
+      completer.complete(value);
+    });
+
+    return _cache[id];
+  }
+
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
+  /// [TValue] resolved from the [Future] returned by factory when the value
+  /// does not exist in the cache.
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
+  /// returning the cached value. If the [Cache] does not contain an instance
+  /// for the given [TIdentifier], the given [valueFactory] is called and an
+  /// [CacheContext] event is emitted on the [didUpdate] stream. A call to [get]
+  /// that returns a cached value does not emit this event as the [Cache] has
+  /// not updated.
+  ///
+  /// Calls to [get] are evaluated syncronously and will return the future value
+  /// established by the first call to [get]. This allows calls to get to
+  /// return the same value regardless of completion of asyncronous calls to
+  /// the [onGet] lifecycle methods or [valueFactory].
+  ///
+  /// ```
+  /// var a = cache.getAsync('id', _superLongAsyncCall);
+  /// var b = cache.getAsync('id', _superLongAsyncCall);
+  /// var c = cache.getAsync('id', _superLongAsyncCall);
+  ///
+  /// // All of the values in this list are the same instance returned by the
+  /// // first call to `_superLongAsyncCall`.
+  /// var values = Future.wait([a, b, c]);
+  /// ```
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  @mustCallSuper
+  Future<TValue> getAsync(TIdentifier id, Func<Future<TValue>> valueFactory) {
+    _throwWhenDisposed('getAsync');
+
+    // Await any pending cached futures
+    if (_cache.containsKey(id)) {
+      return _cache[id].then((TValue value) async {
+        await onGet(id, value);
+        return value;
+      });
+    }
+
+    // Install Future value
+    final Completer<TValue> completer = new Completer<TValue>();
+    _cache[id] = completer.future;
+    valueFactory.call().then((TValue value) async {
+      await onGet(id, value);
+      _didUpdateController.add(new CacheContext(id, value));
+      completer.complete(value);
+    });
+
+    return _cache[id];
+  }
+
+  /// Is the given identifier in the [Cache]?
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  bool isCached(TIdentifier id) {
+    _throwWhenDisposed('determine if id is cached');
+    return _cache.containsKey(id);
+  }
+
+  /// Updates the current value for a given [TIdentifier] with the given
+  /// [TValue].
+  ///
+  /// Putting a value into the [Cache] results in the imeddiate replacement. any
+  /// future call to [get] or [getAsync] will return the given [value]. Any call
+  /// to put will await the completion of the [onPut] lifecycle method before
+  /// returning the given value. A [CacheContext] event is emitted by the
+  /// [didUpdate] stream upon completion of the [put].
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  Future<TValue> put(TIdentifier id, TValue value) {
+    _throwWhenDisposed('put');
+    final Completer<TValue> completer = new Completer<TValue>();
+    _cache[id] = completer.future;
+
+    onPut(id, value).then((Null _) {
+      _didUpdateController.add(new CacheContext(id, value));
+      completer.complete(value);
+    });
+
+    return _cache[id];
+  }
+
+  /// Allows consumers to define behavior that is executed when a [TValue] is
+  /// obtained from the [Cache].
+  @protected
+  Future<Null> onGet(TIdentifier id, TValue value) async {}
+
+  /// Allows consumers to define behavior that is executed when a [TValue] is
+  /// updated in the [Cache].
+  @protected
+  Future<Null> onPut(TIdentifier id, TValue value) async {}
+
+  /// Allows consumers to define behavior that is executed when a [TValue] is
+  /// removed.
+  @protected
+  Future<Null> onRemove(TIdentifier id, TValue value) async {}
+
+  /// Removes the reference to a [TValue] associated with the given
+  /// [TIdentifier].
+  ///
+  /// If a value is currently cached for the given [TIdentifier], the completion
+  /// of the [get] or [put] that inserted the [TIdentifier] into the cache is
+  /// awaited before removal. This ensures that any pending asyncronous calls
+  /// against the cache resovle to a value.
+  ///
+  /// A [CacheContext] value is emitted by the [didUpdate] stream upon
+  /// successfull removal of the given [TIdentifier].
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  @mustCallSuper
+  Future<Null> remove(TIdentifier id) {
+    _throwWhenDisposed('remove');
+
+    if (_cache.containsKey(id)) {
+      return _removalCompleters.putIfAbsent(id, () {
+        final completer = new Completer<Null>();
+
+        _cache[id].then((TValue value) async {
+          await onRemove(id, value);
+          await _cache.remove(id);
+          _removalCompleters.remove(id);
+          _didRemoveController.add(new CacheContext(id, value));
+          _didUpdateController.add(new CacheContext(id, null));
+          completer.complete();
+        });
+
+        return completer;
+      }).future;
+    }
+
+    return new Future.value();
+  }
+
+  void _throwWhenDisposed(String op) {
+    if (isDisposedOrDisposing) {
+      throw new StateError('Cannot $op when Cache is disposed');
+    }
+  }
+}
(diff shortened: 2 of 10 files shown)


aviary-wf:
## Raven
Number of Findings: 0

dustinlessard-wf:
This is very awesome!

README.md
```diff
@@ -3,13 +3,18 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * `Disposable` interface / mixin to assist with cleaning up streams and other
-  data structures that won't necessarily be garbage collected without some
+  * A `Cache` implementation that maintians references to an object instance by
```
dustinlessard-wf: typo : maintians

test/unit/vm/disposable_test.dart
```diff
@@ -89,15 +87,16 @@ void main() {
       test(
           'should call callback and accept null return value'
           'when parent is disposed', () async {
-        thing.testManageDisposer(expectAsync(() => null, count: 1) as Disposer);
+        thing
```
bencampbell-wf: Updated deprecated uses of `expectAsync`.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,237 @@
...
+  /// Removes the reference to a [TValue] associated with the given
+  /// [TIdentifier].
+  ///
+  /// If a value is currently cached for the given [TIdentifier], the completion
+  /// of the [get] or [put] that inserted the [TIdentifier] into the cache is
+  /// awaited before removal. This ensures that any pending asyncronous calls
+  /// against the cache resovle to a value.
+  ///
+  /// A [CacheContext] value is emitted by the [didUpdate] stream upon
+  /// successfull removal of the given [TIdentifier].
```
bencampbell-wf: The documentation should reflect that the didRemove stream is emitted on when a removal succeeds.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,237 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'package:w_common/disposable.dart';
+import 'package:w_common/func.dart';
+
+/// Immutable payload that indicates a change in a [Cache].
+class CacheContext<TIdentifier, TValue> {
+  /// Identifies a single instance of a [TValue]
```
aaronstgeorge-wf: Is there any point it writing brackets around a parameterized type?

bencampbell-wf: It is consistent with not-parameterized types.

aaronstgeorge-wf: For future reference dart doc will format anything surrounded by brackets that it can't link as inline code. 

screen shot 2017-01-04 at 4 14 31 pm

looks good to me 👍

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,237 @@
...
+/// References are retained for the lifecycle of the instance of the [Cache],
+/// unless explicitly removed.
+class Cache<TIdentifier, TValue> extends Object with Disposable {
+  Map<TIdentifier, Future<TValue>> _cache = <TIdentifier, Future<TValue>>{};
+  StreamController<CacheContext<TIdentifier, TValue>> _didUpdateController =
+      new StreamController<CacheContext<TIdentifier, TValue>>.broadcast();
+  StreamController<CacheContext<TIdentifier, TValue>> _didRemoveController =
+      new StreamController<CacheContext<TIdentifier, TValue>>.broadcast();
+  Map<TIdentifier, Completer<Null>> _removalCompleters =
+      <TIdentifier, Completer<Null>>{};
```
aaronstgeorge-wf: These are a little hard to read, some whitespace around them might be nice.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,237 @@
...
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  bool isCached(TIdentifier id) {
+    _throwWhenDisposed('determine if id is cached');
+    return _cache.containsKey(id);
+  }
+
+  /// Updates the current value for a given [TIdentifier] with the given
+  /// [TValue].
+  ///
+  /// Putting a value into the [Cache] results in the imeddiate replacement. any
```
aaronstgeorge-wf: `any` should be capitalized.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,237 @@
...
+  /// future call to [get] or [getAsync] will return the given [value]. Any call
+  /// to put will await the completion of the [onPut] lifecycle method before
+  /// returning the given value. A [CacheContext] event is emitted by the
+  /// [didUpdate] stream upon completion of the [put].
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  Future<TValue> put(TIdentifier id, TValue value) {
```
aaronstgeorge-wf: This method could be below the `on..`s for alphabetization.

test/unit/vm/cache/cache_test.dart
```diff
@@ -0,0 +1,301 @@
...
+        var didCallValueFactory = false;
+        await cache.get(notCachedId, () {
+          didCallValueFactory = true;
+          return notCachedValue;
+        });
+        expect(didCallValueFactory, isTrue);
+      });
+
+      test('should not call valueFactory if id is cached', () async {
+        var didCallValueFactory = false;
```
aaronstgeorge-wf: Looks like this test still needs to be updated.

test/unit/vm/cache/cache_test.dart
```diff
@@ -0,0 +1,301 @@
...
+
+      test(
+          'should dispatch one didRemove event when identifier is removed '
+          'syncronously', () {
+        cache.didRemove.listen(expectAsync1((CacheContext context) {
+          expect(context.id, cachedId);
+          expect(context.value, cachedValue);
+        }, count: 1));
+        cache.remove(cachedId);
+        cache.remove(cachedId);
```
aaronstgeorge-wf: Would a future have to be awaited here? I don't know if this would fail after the test is already finished.

lib/src/cache/cache.dart
bencampbell-wf: This has been addressed.

lib/src/cache/reference_cache.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'package:w_common/src/cache/cache.dart';
+
+/// Maintains the number of references for an instance of an cache value.
+class ReferenceCache<TIdentifier, TValue> extends Cache<TIdentifier, TValue> {
```
bencampbell-wf: I am totally up for a better name. I would like to solicit more feedback and get a list of other suggestions before changing the code.

README.md
```diff
@@ -3,13 +3,18 @@
...
+    * `ReferenceCache` maintains a count for each access for a given identifier.
+    When the final reference is released, the item is removed from the cache.
+  * A `Disposable` interface / mixin to assist with cleaning up streams and
+  other data structures that won't necessarily be garbage collected without some
   manual intervention.
   * A simple typedef that can be parameterized to represent a zero-arity
   callback that returns a particular type.
-  * `InvalidationMixin` mixin used to mark a class as requiring validation.
-  * `JsonSerializable` interface to indicate that something can be serialized
+  * A `InvalidationMixin` mixin used to mark a class as requiring validation.
```
georgelesica-wf: "An"

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'package:w_common/disposable.dart';
+import 'package:w_common/func.dart';
+
+/// Immutable payload that indicates a change in a [Cache].
+class CacheContext<TIdentifier, TValue> {
```
georgelesica-wf: `CachePayload`? `CachedItem`? `CacheResident`? Naming is hard. I don't love "context".

README.md
```diff
@@ -3,13 +3,18 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * `Disposable` interface / mixin to assist with cleaning up streams and other
-  data structures that won't necessarily be garbage collected without some
+  * A `Cache` implementation that maintains references to an object instance by
+  an identifier. Specializations of this class allow invariants to be
+  maintained:
+    * `ReferenceCache` maintains a count for each access for a given identifier.
```
georgelesica-wf: "A"?

lib/src/cache/cache.dart
bencampbell-wf: As suggested above, I will wait till everyone has a chance to weigh in on possible names before making a change.

georgelesica-wf: Hold off, I might have misunderstood how this was used (because I hadn't read the rest of the code yet...)

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// Identifies a single instance of a [TValue]
+  final TIdentifier id;
+
+  /// The current value stored in the [Cache] for the attributed [TIdentifier].
+  final TValue value;
+
+  CacheContext(this.id, this.value);
+}
+
+/// Maintians a reference to a given any [TValue] attributed to a [TIdentifier].
```
georgelesica-wf: "to a given any [TValue]"? Remove the "any" maybe?

georgelesica-wf: "attributed" might not be quite the right word here. Maybe "tied"? "connected"? "associated with"?

georgelesica-wf: Also, "maintians".

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  // ignore: close_sinks
+  StreamController<CacheContext<TIdentifier, TValue>> _didRemoveController =
+      new StreamController<CacheContext<TIdentifier, TValue>>.broadcast();
+
+  /// A collection of pending removals for a given [TIdentifier].
+  Map<TIdentifier, Completer<Null>> _removalCompleters =
+      <TIdentifier, Completer<Null>>{};
+
+  Cache() {
+    [_didRemoveController, _didUpdateController]
```
georgelesica-wf: #nit alpha order would be nicer

georgelesica-wf: Actually, these *are* in the right order, their declarations above are not...

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+        .forEach(manageStreamController);
+  }
+
+  /// The collection of removals from the [Cache].
```
georgelesica-wf: "collection"? It's a stream of events, right?

lib/src/cache/cache.dart
bencampbell-wf: I'm confused. `didR*` < `didU*`,  correct?

lib/src/cache/cache.dart
sebastianmalysa-wf: "Maintains... and "...to a given [TValue] attributed..." or "...to any given [TValue]"

lib/src/cache/cache.dart
bencampbell-wf: I try not to restate types in doc comments. Collection has been my goto for `Stream`s as function like indefinite lists. I can change it if it adds confusion.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  Stream<CacheContext<TIdentifier, TValue>> get didRemove =>
+      _didRemoveController.stream;
+
+  /// The collection of changes to the [Cache].
+  ///
+  /// Values are emitted when the cached value is changes.
```
sebastianmalysa-wf: "...changed."

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
+  /// returning the cached value. If the [Cache] does not contain an instance
+  /// for the given [TIdentifier], the given [valueFactory] is called and an
+  /// [CacheContext] event is emitted on the [didUpdate] stream. A call to [get]
+  /// that returns a cached value does not emit this event as the [Cache] has
+  /// not updated.
+  ///
+  /// Calls to [get] are evaluated syncronously and will return the future value
+  /// established by the first call to [get]. This allows calls to get to
```
sebastianmalysa-wf: Should the "get" here "...This allows calls to get..." be in brackets?

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
+      _didUpdateController.stream;
+
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
+  /// [TValue] returned by the value factory when the value does not exist in
+  /// the cache.
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
+  /// returning the cached value. If the [Cache] does not contain an instance
+  /// for the given [TIdentifier], the given [valueFactory] is called and an
```
sebastianmalysa-wf: "...is called and a..."

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+
+  /// The collection of changes to the [Cache].
+  ///
+  /// Values are emitted when the cached value is changes.
+  Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
+      _didUpdateController.stream;
+
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
+  /// [TValue] returned by the value factory when the value does not exist in
+  /// the cache.
```
georgelesica-wf: It adds *and* returns the value from the value factory if the identifier wasn't found, right? Maybe clarify this.

lib/src/cache/cache.dart
georgelesica-wf: Yeah, but on lines 28 and 32 you declare them in the reverse order. Sorry, my original comment was confusing, the alphabet is hard.

lib/src/cache/cache.dart
georgelesica-wf: I think we generally talk about things like this as event streams, so I'd prefer that nomenclature. The term "collection" is super overloaded, so to me it's less clear despite being technically correct.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// ```
+  /// var a = cache.getAsync('id', _superLongAsyncCall);
+  /// var b = cache.getAsync('id', _superLongAsyncCall);
+  /// var c = cache.getAsync('id', _superLongAsyncCall);
+  ///
+  /// // All of the values in this list are the same instance returned by the
+  /// // first call to `_superLongAsyncCall`.
+  /// var values = Future.wait([a, b, c]);
+  /// ```
+  ///
```
sebastianmalysa-wf: The above documentation repeats a lot, was this the intention?

lib/src/cache/cache.dart
georgelesica-wf: As a more general fix you might consider using "this method" instead of "[get]" in some cases.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+
+  /// Allows consumers to define behavior that is executed when a [TValue] is
+  /// removed.
+  @protected
+  Future<Null> onRemove(TIdentifier id, TValue value) async {}
+
+  /// Updates the current value for a given [TIdentifier] with the given
+  /// [TValue].
+  ///
+  /// Putting a value into the [Cache] results in the imeddiate replacement. Any
```
sebastianmalysa-wf: "...immediate..."

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
+  /// [TValue] returned by the value factory when the value does not exist in
+  /// the cache.
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
+  /// returning the cached value. If the [Cache] does not contain an instance
+  /// for the given [TIdentifier], the given [valueFactory] is called and an
+  /// [CacheContext] event is emitted on the [didUpdate] stream. A call to [get]
+  /// that returns a cached value does not emit this event as the [Cache] has
+  /// not updated.
```
georgelesica-wf: "not changed."?

lib/src/cache/cache.dart
bencampbell-wf: I thought it was better to be repetitive so a user could understand the nuances of `get` and `getAsync` without having to look beyond the doc comment of each function.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// [TIdentifier].
+  ///
+  /// If a value is currently cached for the given [TIdentifier], the completion
+  /// of the [get] or [put] that inserted the [TIdentifier] into the cache is
+  /// awaited before removal. This ensures that any pending asyncronous calls
+  /// against the cache resovle to a value.
+  ///
+  /// A [CacheContext] value is emitted by the [didUpdate] stream upon
+  /// successfull removal of the given [TIdentifier]. In addition, a
+  /// [CacheContext] value is emitted by the [didRemove] stream upon successfull
```
sebastianmalysa-wf: "...successful..." there is one above this line as well :)

lib/src/cache/reference_cache.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart';
+
+import 'package:w_common/src/cache/cache.dart';
+
+/// Maintains the number of references for an instance of an cache value.
```
sebastianmalysa-wf: "... a cache value."

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// return the same value regardless of completion of asyncronous calls to
+  /// the [onGet] lifecycle methods or [valueFactory].
+  ///
+  /// ```
+  /// var a = cache.get('id', _superLongCall);
+  /// var b = cache.get('id', _superLongCall);
+  /// var c = cache.get('id', _superLongCall);
+  ///
+  /// // All of the values in this list are the same instance returned by the
+  /// // first call to `_superLongCall`.
```
georgelesica-wf: "first call"... there *was* only one call, right? I think a better way to state this would be to say something like... "The futures in this list will all resolve to the same instance, which was returned by `_superLongCall` the first time `get` was called."

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+  /// var values = Future.wait([a, b, c]);
+  /// ```
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
```
georgelesica-wf: My preference would be to just write out "is disposed or disposing".

bencampbell-wf: I went with the reference as it allows consumers to jump to the definition which is explicit in what that reference means.

lib/src/cache/cache.dart
bencampbell-wf: Everything will be alphabetical order.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+
+    onGet(id, value).then((Null _) {
+      _didUpdateController.add(new CacheContext(id, value));
+      completer.complete(value);
+    });
+
+    return _cache[id];
+  }
+
+  /// Returns a value from the cache for a given [TIdentifier] or adds the
```
georgelesica-wf: "adds and returns" or whatever

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+      completer.complete(value);
+    });
+
+    return _cache[id];
+  }
+
+  /// Is the given identifier in the [Cache]?
+  ///
+  /// If the [Cache] [isDisposedOrDisposing] then a [StateError] is thrown.
+  bool isCached(TIdentifier id) {
```
georgelesica-wf: It might be better to just call this `containsKey` to match up with `Map` better.

lib/src/cache/cache.dart
georgelesica-wf: I would actually prefer to see the documentation defined in terms of the [get] method since they are virtually identical. Having the full documentation basically copy-pasted obscures the purpose of this method and makes it more likely that the two will get out of sync.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+    _throwWhenDisposed('determine if identifier is cached');
+    return _cache.containsKey(id);
+  }
+
+  /// Allows consumers to define behavior that is executed when a [TValue] is
```
georgelesica-wf: For these callbacks I would just say something like "Logic executed when..." instead of the bit about allowing the consumer to do something. See https://github.com/Workiva/w_module/blob/master/lib/src/lifecycle_module.dart#L452 for an example.

lib/src/cache/cache.dart
georgelesica-wf: Also, "the" is either unnecessary or you need something after "replacement"... "results in the immediate replacement of the existing value" (or whatever).

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,245 @@
...
+    return _cache[id];
+  }
+
+  /// Removes the reference to a [TValue] associated with the given
+  /// [TIdentifier].
+  ///
+  /// If a value is currently cached for the given [TIdentifier], the completion
+  /// of the [get] or [put] that inserted the [TIdentifier] into the cache is
+  /// awaited before removal. This ensures that any pending asyncronous calls
+  /// against the cache resovle to a value.
```
georgelesica-wf: "resolve"

lib/src/cache/reference_cache.dart
georgelesica-wf: "to an instance"

README.md
```diff
@@ -3,13 +3,18 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * `Disposable` interface / mixin to assist with cleaning up streams and other
-  data structures that won't necessarily be garbage collected without some
+  * A `Cache` implementation that maintains references to an object instance by
+  an identifier. Specializations of this class allow invariants to be
+  maintained:
+    * A `ReferenceCache` maintains a count for each access for a given identifier.
+    When the final reference is released, the item is removed from the cache.
```
todbachman-wf: This renders a little awkwardly and probably not how you intended. It might be nice if this was indented in the rendered markup. This is how it looks now:

image

bencampbell-wf: Yikes, good catch! That is not how I expected it to render at all.

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,241 @@
...
+
+  /// The stream of events that indicate a change to a cached [TValue].
+  Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
+      _didUpdateController.stream;
+
+  /// Returns a value from the cache for a given [TIdentifier]. If the
+  /// [TIdentifier] is not present in the cache the value returned by the given
+  /// [valueFactory] is added to the cache and returned.
+  ///
+  /// All calls to [get] await a call to the [onGet] lifecycle method before
```
georgelesica-wf: Just to get this straight in my head, if I call `get` twice in a row with the same identifier without awaiting the resulting future, will `onGet` be called once or twice?

lib/src/cache/cache.dart
```diff
@@ -0,0 +1,241 @@
...
+  /// A collection of pending removals for a given [TIdentifier].
+  Map<TIdentifier, Completer<Null>> _removalCompleters =
+      <TIdentifier, Completer<Null>>{};
+
+  Cache() {
+    [_didRemoveController, _didUpdateController]
+        .forEach(manageStreamController);
+  }
+
+  /// The stream of events that indicate a removal [Cache].
```
todbachman-wf: Maybe?

```Dart
/// A stream of [CacheContext] that dispatches when an item is added to the cache.
```

lib/src/cache/cache.dart
bencampbell-wf: Twice.

georgelesica-wf: Ah, OK, maybe change the words "All calls" to "Each call" just to make that a little clearer?

bencampbell-wf:
@georgelesica-wf suggested implementing this as a separate package as it has grown outside the typical scope and size of a w_common library.  

This seems like a fair suggestion as a Cache library can (and probably will) expand in functionality and complexity.  What are the reviewers thoughts on this?

README.md
```diff
@@ -3,13 +3,17 @@
 A collection of helpful utilities for use in Dart projects. Right now, it
 includes the following:
 
-  * `Disposable` interface / mixin to assist with cleaning up streams and other
-  data structures that won't necessarily be garbage collected without some
+  * A `Cache` implementation that maintains references to an object instance by
+  an identifier. Specializations of this class allow invariant to be
```
dustinlessard-wf: typo maybe? Should `invariant` be plural?

evanweible-wf:
@bencampbell-wf @georgelesica-wf It does feel a bit big to be going into w_common.. I wouldn't be opposed to breaking this out into a separate repo (private for now)

maxwellpeterson-wf:
I think I disagree that this is too big for w_common. If there is concern that this will introduce a significantly higher likelihood of needing breaking changes of this library in the future I could see an argument for separating it. But this seems like a pretty solid, complete implementation. 

I'm not completely opposed, but we should be careful not to introduce extra administrative overhead for minimal gain.

bencampbell-wf:
I am going to close this PR for now and pull this implementation into the repo I am utilizing and hide it from the public library interface. I've created RAP-1516 to capture finding a permanent home for the implementation at a later date.

PR #20, opened 2017-01-09
RAP-1514 Improve manage methods and add manageFuture

georgelesica-wf:
### Description

If we upgrade to Dart 1.21 we get parametric methods, like so:

```dart
T myMethod<T>(T aThing) {
  // ...
}
```

This lets us improve the manage methods be returning their arguments, meaning they can be inserted into existing code more cleanly:

```dart
_controller = manageStreamController(new StreamController.broadcast());
```

We'd also like to be able to automatically await a set of futures before disposing to prevent exceptions when there are outstanding async operations that will access managed sub-objects upon completion. For example, a network request might dispatch a stream controller when it returns.

```dart
manageFuture(_longRunningAsyncThing());
dispose(); // The dispose will wait until the future above completes.
```

The idea here is to allow things like network requests, the callbacks for which might try to access managed assets (like `Disposable`s) on the object, to finish before we start blowing everything away.

### Changes

  * Add `manageFuture` method
  * Update manage methods to be parametric and return their arguments
  * Improve and simplify tests around `Disposable`

Note: we're adding a method to the `DisposableManager` interface, which means that this is technically a breaking change and will require a major version bump.

Note: we need Dart 1.21 for this to work. We may want to start a "2.0-wip" branch so we can release only once consumers are ready for this.

### Semantic Versioning

- [ ] **Patch**
- [ ] **Minor**
* [x] Major
  - [x] This change modifies part of the public API in a backwards-incompatible manner

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf
@travissanderson-wf 
@bencampbell-wf


aviary-wf:
## Raven
Number of Findings: 0

lib/disposable.dart
```diff
@@ -179,16 +182,19 @@ class Disposable implements _Disposable, DisposableManager {
...
       ..add(onDispose());
 
     _internalDisposables = [];
+    _blockingFutures = new Set<Future>();
 
-    // We need to filter out nulls because a subscription cancel
-    // method is allowed to return a plain old null value.
     return Future
-        .wait(futures.where((future) => future != null))
+        .forEach(futures, (future) => future)
```
greglittlefield-wf: `.wait(futures)` ?

georgelesica-wf: They need to be ordered. I ended up disliking the `forEach` version because it was ugly. New version is much easier to understand.

lib/disposable.dart
```diff
@@ -179,63 +182,83 @@ class Disposable implements _Disposable, DisposableManager {
...
   }
 
+  /// Causes the object to add the future to a list of futures which will
+  /// be awaited before the object disposes itself or any of the objects
+  /// under its management.
+  @mustCallSuper
+  @override
+  Future<T> manageFuture<T>(Future<T> future) {
+    _throwOnInvalidCall(future, 'manageFuture');
+    if (!_blockingFutures.contains(future)) {
```
greglittlefield-wf: Won't this check always pass, since .then returns a new future?

georgelesica-wf: Yeah, good catch. I changed the inner code at some point and didn't think about the check.

lib/disposable.dart
```diff
@@ -179,63 +182,83 @@ class Disposable implements _Disposable, DisposableManager {
...
   @mustCallSuper
   @override
-  void manageDisposer(Disposer disposer) {
-    _throwIfNull(disposer, 'disposer');
+  void manageDisposer<T>(Disposer disposer) {
+    _throwOnInvalidCall(disposer, 'disposer');
     _internalDisposables.add(new _InternalDisposable(disposer));
   }
 
+  /// Causes the object to add the future to a list of futures which will
```
greglittlefield-wf: #nit s/Causes the object to add the future/Adds [future]/

lib/disposable.dart
```diff
@@ -179,63 +182,83 @@ class Disposable implements _Disposable, DisposableManager {
...
+  /// be awaited before the object disposes itself or any of the objects
+  /// under its management.
+  @mustCallSuper
+  @override
+  Future<T> manageFuture<T>(Future<T> future) {
```
greglittlefield-wf: Is `manageFuture` really the best name for this?

Seems semantically different then the other methods—this disposable isn't really performing anything on the future.

What about something like `wait`?

greglittlefield-wf: Or `block`, to match the `_blockingFutures` name you've chosen

georgelesica-wf: Yeah, I had originally named it `awaitFutureBeforeDispose` or something like that. I'm open to a better name. I'm not sold on something as simple as `wait`, though, since I think that would suggest something more like `Future.wait`.

lib/disposable.dart
georgelesica-wf: Fixed.

lib/disposable.dart
```diff
@@ -179,63 +182,86 @@ class Disposable implements _Disposable, DisposableManager {
...
+    _throwOnInvalidCall(future, 'manageFuture');
+    if (!_blockingFutures.contains(future)) {
+      Future removeFuture;
+      removeFuture = future.then((_) {
+        _blockingFutures.remove(future);
+        _blockingFutures.remove(removeFuture);
```
greglittlefield-wf: ~~Is it really necessary to remove this? Why not just use the future itself?~~

Okay, I think I see now. This tripped me up a bit, though; might be good to add a code comment.

I assume you're not using just the future itself so that you can "release" it as soon as possible, is that correct?

georgelesica-wf: I want to make sure the "release" operation is also awaited, otherwise the list itself might have been nulled out at the point where it tried to remove itself. I'll add a comment.

greglittlefield-wf: Sorry, by "release", I meant remove the reference to the original Future, allowing it to be garbage collected—not removing the future and its removeFuture from the list.

pubspec.yaml
```diff
@@ -5,15 +5,15 @@ author: Workiva Client Platform Team <team-clientplatform@workiva.com>
 homepage: https://www.github.com/Workiva/w_common
 
 environment:
-  sdk: '>=1.12.0 <2.0.0'
+  sdk: '>=1.21.0 <2.0.0'
```
travissanderson-wf: that's a hefty jump :P

lib/disposable.dart
georgelesica-wf: After side-channel discussion, your understanding is spot-on. For posterity: there could be many, many futures being generated and waited on so I wanted to make sure they can be GC'd.

pubspec.yaml
georgelesica-wf: Truth. That's partly why I didn't make these changes sooner.

georgelesica-wf:
One problem I see here is that consumers will still need to guard their async calls with something like

```dart
if (!isDisposedOrDisposing) {
  waitBeforeDispose(asyncCall());
}
```

because attempting to call `waitBeforeDispose` (or any of the other "manage" methods) after `dispose` has been called will throw.

I don't see a way around this, however. We can't let it go silently because then the async callbacks might throw, just as currently happens.

lib/disposable.dart
```diff
@@ -217,25 +223,29 @@ class Disposable implements _Disposable, DisposableManager {
...
   /// Automatically cancel a stream subscription when this object is disposed.
   ///
   /// The parameter may not be `null`.
   @mustCallSuper
   @override
-  void manageStreamSubscription(StreamSubscription subscription) {
-    _throwIfNull(subscription, 'subscription');
+  StreamSubscription<T> manageStreamSubscription<T>(
+      StreamSubscription<T> subscription) {
+    _throwOnInvalidCall(subscription, 'subscription');
```
evanweible-wf: I think `'subscription'` should be `'manageStreamSubscription'` here, right?

test/unit/browser/invalidation_mixin_test.dart
```diff
@@ -49,12 +49,12 @@ void main() {
...
 
         // ignore: unawaited_futures
-        onValidation.then(expectAsync((ValidationStatus status) {
+        onValidation.then(expectAsync1((ValidationStatus status) {
           expect(status, equals(ValidationStatus.complete));
         }, count: 1));
 
         // ignore: STRONG_MODE_DOWN_CAST_COMPOSITE
-        thing.onValidate.listen(expectAsync((_) {}, count: 1));
+        thing.onValidate.listen(expectAsync1((_) {}, count: 1));
```
evanweible-wf: Do you still need the `ignore` rule above? I thought these new `expectAsync` methods were supposed to obviate their need

lib/disposable.dart
```diff
@@ -143,10 +145,13 @@ typedef Future<dynamic> Disposer();
 /// without explicit reference to [Disposable]. To do this, we use
 /// composition to include the [Disposable] machinery without changing
 /// the public interface of our class or polluting its lifecycle.
+///
+/// Implementing the [DisposableManager] interface is not required.
 class Disposable implements _Disposable, DisposableManager {
   Completer<Null> _didDispose = new Completer<Null>();
   bool _isDisposing = false;
   List<_Disposable> _internalDisposables = [];
+  Set<Future> _blockingFutures = new Set<Future>();
```
bencampbell-wf: Will strong mode complain if `Future` is missing a type, ie this type signature is `Set<Future<dynamic>>`? Would it better to keep a set of null valued futures instead and wrap all pending calls in async wrapper function that gurantees it meets this type instead?

```dart
Set<Future<Null>> _blockingFutures = new Set<Future<Null>>();

/// Other good codes...

Future<T> waitBeforeDispose<T>(Future<T> future) {
     _throwOnInvalidCall(future, 'waitBeforeDispose');
     if (!_blockingFutures.contains(future)) {
       Future<Null> removeFuture;
       Future<Null> opFuture = new Future<Null>(() async {
           await future;
        });
       removeFuture = opFuture.then((T _) {
         _blockingFutures.remove(opFuture);
         _blockingFutures.remove(removeFuture);
      });

      _blockingFutures.addAll([opFuture, removeFuture]);
      return future;
}
```

This is more for explicit type safety which may be worth the indirection for a library that functions with a high degree of type flexibility.

georgelesica-wf: I don't think there's anything wrong with using `Future<dynamic>` in this case, it is still correctly typed since we ignore the value it resolves to. I can be more explicit about it though.

evanweible-wf:
+1 looks great!

bencampbell-wf:
+1

maxwellpeterson-wf:
+1

todbachman-wf:
+10. Confirmed CI passed.

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - N/A
- [x] All unit tests pass

Merging into master.

PR #22, opened 2017-02-23
RAP-1654 Add Managed Timer Factories

bencampbell-wf:
### Description

There are instances where timers are used within a disposable context. These timers can live beyond a call to `dispose` and will execute it's callback function against a disposed object.

### Changes

Two factory methods are added to provide consumers with a way to create Timers that will be cancelled on disposal. An internal implementation of `Timer` is provided which provides `Future` that completes when either the callback of a non-periodic timer is finished or if any type of timer is cancelled. This ensures that any concluded timer is removed from the internal disposable collection.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
  - [x] This change deprecates a part of the public API
* [ ] Major

### Testing/QA

- [ ] CI passes
- [ ] Run analyzer in strong mode to confirm the only hints/lints are for deprecated interface.

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf


aviary-wf:
## Raven
Number of Findings: 0

dustinlessard-wf:
+1, solid

lib/disposable.dart
```diff
@@ -192,29 +215,46 @@ class Disposable implements _Disposable, DisposableManager {
...
+    _internalDisposables.add(disposable);
+
+    return timer;
+  }
+
+  /// Creates a periodic [Timer] that will be cancelled if active upon disposal.
+  @mustCallSuper
+  @override
+  Timer getManagedPeriodicTimer(Duration duration, void callback(Timer timer)) {
+    var timer = new Timer.periodic(duration, callback);
```
todbachman-wf: What will happen if the consumer cancels this timer?

bencampbell-wf: I thought the same thing. Cancelling a timer is idempotent so disposing a cancelled timer will have no effect. The only drawback is that we hold onto the the timer for the life of the Disposable object. I am going to implement `Timer` and add a lifecycle hook for when the timers complete or are cancelled.

bencampbell-wf:
Given this is a breaking change I can retarget the PR at a 2.0.0-wip. My only concern is that `w_module` is dependent on `^1.0.0` and it is non trivial to upgrade consumers of it.

todbachman-wf:
+1. 👏

lib/disposable.dart
```diff
@@ -37,6 +37,43 @@ class _InternalDisposable implements _Disposable {
...
+/// cancelled.
+class _ObservableTimer implements Timer {
+  Completer<Null> _didConclude = new Completer<Null>();
+  Timer _timer;
+
+  _ObservableTimer(Duration duration, void callback()) {
+    _timer = new Timer(duration, () {
+      callback();
+
+      if (!_didConclude.isCompleted) {
```
georgelesica-wf: Would we want to put the call to the callback in the guarded block as well? Presumably we would never want the callback to fire if the future has already completed. No?

lib/disposable.dart
```diff
@@ -192,29 +250,38 @@ class Disposable implements _Disposable, DisposableManager {
...
+  /// Creates a [Timer] instance that will be cancelled if active upon disposal.
+  @mustCallSuper
+  @override
+  Timer getManagedTimer(Duration duration, void callback()) {
+    var timer = new _ObservableTimer(duration, callback);
+    _addObservableTimerDisposable(timer);
+    return timer;
+  }
+
+  /// Creates a periodic [Timer] that will be cancelled if active upon disposal.
```
georgelesica-wf: You probably don't want doc strings here since they're on the `DisposableManager` abstract class, right?

test/unit/browser/invalidation_mixin_test.dart
```diff
@@ -36,9 +36,9 @@ void main() {
         expect(thing.invalid, isTrue);
 
         // ignore: unawaited_futures
-        onValidation.then(expectAsync((ValidationStatus status) {
+        onValidation.then(expectAsync1((ValidationStatus status) {
           expect(status, equals(ValidationStatus.cancelled));
-        }, count: 1));
```
georgelesica-wf: I like being explicit with the counts when the count is largely what is being tested, but if people don't like it then that's fine.

test/unit/stubs.dart
```diff
@@ -0,0 +1,112 @@
...
+
+import 'package:test/test.dart';
+import 'package:w_common/disposable.dart';
+
+import 'typedefs.dart';
+
+class DisposableThing extends Object with Disposable {
+  bool wasOnDisposeCalled = false;
+
+  void testManageDisposable(Disposable thing) {
```
georgelesica-wf: If you're feeling ambitious, these `test*` methods exist because the `manage*` methods used to be `@protected`, which is no longer the case. So you could remove them and update the tests.

test/unit/vm/disposable_test.dart
```diff
@@ -63,6 +26,60 @@ void main() {
...
+      });
+
+      test('should cancel timer if disposed before completion', () async {
+        expect(timer.isActive, isTrue);
+        await thing.dispose();
+        expect(await harness.didCancelTimer, isTrue);
+        expect(await harness.didCompleteTimer, isFalse);
+      });
+
+      test('disposing should have no effect on timer after it has completed',
```
georgelesica-wf: I didn't think periodic timers were supposed to conclude on their own...

test/unit/browser/invalidation_mixin_test.dart
bencampbell-wf: Ha. @todbachman-wf suggested in a previous PR that I remove redundant counts. I obliged.

lib/disposable.dart
bencampbell-wf: This guard is unnecessary in reality. The underlying `Timer` implementation would never call this function if it were cancelled.

test/unit/vm/disposable_test.dart
bencampbell-wf: I updated the test description to indicate that when a consumer cancels the timer a dispose has no effect.

lib/src/disposable/disposable_manager.dart
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// Managers for disposable members.
+///
+/// This interface allows consumers to exercise more control over how
+/// disposal is implemented for their classes.
+///
+/// When new management methods are to be added, they should be added
+/// here first, then implemented in [Disposable].
+abstract class DisposableManagerV2 implements DisposableManager {
```
evanweible-wf: I don't think you need to repeat the methods that are already defined in V1 (and not changed in V2)

test/unit/browser/invalidation_mixin_test.dart
```diff
@@ -49,12 +49,12 @@ void main() {
...
         // ignore: unawaited_futures
-        onValidation.then(expectAsync((ValidationStatus status) {
+        onValidation.then(expectAsync1((ValidationStatus status) {
           expect(status, equals(ValidationStatus.complete));
-        }, count: 1));
+        }));
 
         // ignore: STRONG_MODE_DOWN_CAST_COMPOSITE
-        thing.onValidate.listen(expectAsync((_) {}, count: 1));
+        thing.onValidate.listen(expectAsync1((_) {}));
```
evanweible-wf: are you able to get rid of the `// ignore: STRONG_MODE_DOWN_CAST_COMPOSITE` with this change?

bencampbell-wf: I am!

lib/src/disposable/disposable_manager.dart
bencampbell-wf: You are correct.  I have removed the overridden members.

georgelesica-wf:
+1

todbachman-wf:
+10

#### Testing Performed:
- [x] CI passes
- [x] Run analyzer in strong mode to confirm the only hints/lints are for deprecated interface.

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - N/A
- [x] All unit tests pass

@Workiva/release-management-p for merge into master

bencampbell-wf:
## Release Notes

Updates the `DisposableManagement` interface to provide two factory functions for obtaining timers that will be canceled upon disposal. Consumers should update any instances that implement the `DisposableManagement` interface to implement the 'DisposableManagementV2` interface.

PR #24, opened 2017-03-17
RAP-1716 Add some additional management and helper methods

georgelesica-wf:
### Description

Based on some of our discussions around memory leaks and best practices I decided to add a few more, hopefully helpful utility methods.

### Changes

Add the following:

  * `Future<T> awaitBeforeDispose<T>(Future<T> future)` - causes the object to await the given future before disposing. The future itself is returned for convenience.
  * `Future<T> getManagedDelayedFuture<T>(Duration duration, T callback())` - returns a delayed future that will complete with the result of calling `callback` after a specific duration. If the object is disposed before the duration has elapsed the future will still complete, but with an `ObjectDisposedException` error.
  * `Completer<T> manageCompleter<T>(Completer<T> completer)` - manages a completer by completing it with an `ObjectDisposedException` error if the object is disposed before the completer has completed.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
  - [x] This change deprecates a part of the public API
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf
@Workiva/rich-app-platform-pp

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -34,7 +34,6 @@ linter:
     - library_names
     - library_prefixes
     - non_constant_identifier_names
-    - one_member_abstracts
     - only_throw_errors
     - overridden_fields
     - package_api_docs
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -232,7 +232,16 @@ class Disposable implements _Disposable, DisposableManagerV3 {
   @mustCallSuper
   @override
   Future<T> getManagedDelayedFuture<T>(Duration duration, T callback()) {
-    return null;
+    var completer = new Completer<T>();
+    var timer =
+        new _ObservableTimer(duration, () => completer.complete(callback()));
+    var disposable = new _InternalDisposable(() async {
+      timer.cancel();
+      completer.completeError(new ObjectDisposedException());
+    });
+    _internalDisposables.add(disposable);
+    timer.didConclude.then((Null _) => _internalDisposables.remove(disposable));
+    return completer.future;
   }
 
   @mustCallSuper
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -164,10 +164,11 @@ typedef Future<dynamic> Disposer();
 /// without explicit reference to [Disposable]. To do this, we use
 /// composition to include the [Disposable] machinery without changing
 /// the public interface of our class or polluting its lifecycle.
-class Disposable implements _Disposable, DisposableManagerV2 {
+class Disposable implements _Disposable, DisposableManagerV3 {
+  List<Future<dynamic>> _awaitableFutures = <Future<dynamic>>[];
   Completer<Null> _didDispose = new Completer<Null>();
+  List<_Disposable> _internalDisposables = <_Disposable>[];
   bool _isDisposing = false;
-  List<_Disposable> _internalDisposables = [];
 
   /// A [Future] that will complete when this object has been disposed.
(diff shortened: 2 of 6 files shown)


aviary-wf:
## Raven
Number of Findings: 0

analysis_options.yaml
```diff
@@ -34,7 +34,6 @@ linter:
     - library_names
     - library_prefixes
     - non_constant_identifier_names
-    - one_member_abstracts
```
georgelesica-wf: Removed because I find it annoying and sometimes one-member abstracts are useful as interfaces.

lib/src/disposable/disposable.dart
```diff
@@ -164,10 +164,11 @@ typedef Future<dynamic> Disposer();
 /// without explicit reference to [Disposable]. To do this, we use
 /// composition to include the [Disposable] machinery without changing
 /// the public interface of our class or polluting its lifecycle.
-class Disposable implements _Disposable, DisposableManagerV2 {
+class Disposable implements _Disposable, DisposableManagerV3 {
+  List<Future<dynamic>> _awaitableFutures = <Future<dynamic>>[];
```
bencampbell-wf: This name is kind of goofy. Aren't all `Future`s await-able 😄

lib/src/disposable/disposable.dart
```diff
@@ -189,6 +190,19 @@ class Disposable implements _Disposable, DisposableManagerV2 {
   /// and will become `false` once the [didDispose] future completes.
   bool get isDisposing => _isDisposing;
 
+  @mustCallSuper
+  @override
+  Future<T> awaitBeforeDispose<T>(Future<T> future) {
```
bencampbell-wf: Isn't this the same as: `manageDisposer(() => future);`?

lib/src/disposable/disposable.dart
```diff
@@ -200,12 +214,14 @@ class Disposable implements _Disposable, DisposableManagerV2 {
     }
     _isDisposing = true;
 
-    List<Future> futures = []
+    List<Future<dynamic>> futures = []
```
bencampbell-wf: 🌵 `var futures = <Future<dynamic>>[];`

test/unit/vm/disposable_test.dart
```diff
@@ -103,44 +124,112 @@ void main() {
...
+    });
+
+    group('manageCompleter', () {
+      test('should complete with an error when parent is disposed', () {
+        var completer = new Completer<Null>();
+        completer.future.catchError(expectAsync1((exception) {
+          expect(exception, new isInstanceOf<ObjectDisposedException>());
+        }));
+        thing.manageCompleter(completer);
+        thing.dispose();
```
bencampbell-wf: Should `dispose` return any error it encountered, such as the one above? A consume disposing something may want to handle a dirty disposal.

bencampbell-wf:
+1

lib/src/disposable/disposable.dart
georgelesica-wf: The difference is that the list is tidied up in this case, using `manageDisposer` would accumulate incomplete futures forever.

todbachman-wf:
+10. Confirmed CI passed.

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - N/A
- [x] All unit tests pass

Merging into master

PR #28, opened 2017-04-05
RAP-1780 Handle single-subscription stream controllers better

georgelesica-wf:
### Description

There are two situations in which managing a single-subscription stream controller will result in an exception on dispose.

  1. The stream has already been closed
  2. The stream had a listener and that listener was subsequently canceled *before* disposal occurred

### Changes

Use some hackery to figure out if the stream was listened-to and the listener subsequently canceled its subscription. Also, check to see if the stream was already closed before we try to listen to it again.

As an aside, the reason we listen to the stream before closing the controller is that closing a controller that has never had a listener will cause an exception to be thrown.

### Semantic Versioning

- [x] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

We probably want @dustinlessard-wf and / or @bencampbell-wf to give this a glance to make sure I haven't introduced any new memory leaks.

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### FYI

@seanburke-wf 
@bencampbell-wf

### DIFF
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -30,10 +30,10 @@ class _InternalDisposable implements _Disposable {
 
   @override
   Future<Null> dispose() {
-    var disposeFuture = _disposer();
+    var disposeFuture = _disposer != null ? _disposer() : null;
     _disposer = null;
     if (disposeFuture == null) {
-      return new Future(() => null);
+      return new Future.value();
     }
     return disposeFuture.then((_) => null);
   }
@@ -304,20 +304,22 @@ class Disposable implements _Disposable, DisposableManagerV3 {
     // subscription is subsequently canceled, the `done` future will
     // complete, but there is no other way for us to tell that this
     // is what has happened. If we then listen to the stream (since
-    // closing a stream that was never listen-to throws) we'll get
-    // an exception. This workaround allows us to "know" when a
+    // closing a stream that was never listened to never completes) we'll
+    // get an exception. This workaround allows us to "know" when a
     // subscription has been canceled so we don't bother trying to
     // listen to the stream before closing it.
     bool isDone = false;
-    controller.done.then((_) {
-      isDone = true;
-    });
-    _internalDisposables.add(new _InternalDisposable(() {
+    var disposable = new _InternalDisposable(() {
       if (!controller.hasListener && !controller.isClosed && !isDone) {
         controller.stream.listen((_) {});
       }
       return controller.close();
-    }));
+    });
+    controller.done.then((_) {
+      isDone = true;
+      disposable.dispose();
+    });
+    _internalDisposables.add(disposable);
   }
 
   @mustCallSuper
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -300,8 +300,20 @@ class Disposable implements _Disposable, DisposableManagerV3 {
   @override
   void manageStreamController(StreamController controller) {
     _throwOnInvalidCall('manageStreamController', 'controller', controller);
+    // If a single-subscription stream has a subscription and that
+    // subscription is subsequently canceled, the `done` future will
+    // complete, but there is no other way for us to tell that this
+    // is what has happened. If we then listen to the stream (since
+    // closing a stream that was never listen-to throws) we'll get
+    // an exception. This workaround allows us to "know" when a
+    // subscription has been canceled so we don't bother trying to
+    // listen to the stream before closing it.
+    bool isDone = false;
+    controller.done.then((_) {
+      isDone = true;
+    });
     _internalDisposables.add(new _InternalDisposable(() {
-      if (!controller.hasListener) {
+      if (!controller.hasListener && !controller.isClosed && !isDone) {
         controller.stream.listen((_) {});
       }
       return controller.close();
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -317,6 +317,7 @@ class Disposable implements _Disposable, DisposableManagerV3 {
     });
     controller.done.then((_) {
       isDone = true;
+      _internalDisposables.remove(disposable);
       disposable.dispose();
     });
     _internalDisposables.add(disposable);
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

georgelesica-wf:
## Release Notes

Fixes a bug in which managing a single-subscription stream controller would cause an exception in some cases when the object was later disposed. Managing single-subscription stream controllers should now "just work".

lib/src/disposable/disposable.dart
```diff
@@ -300,8 +300,20 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
+    bool isDone = false;
+    controller.done.then((_) {
+      isDone = true;
+    });
     _internalDisposables.add(new _InternalDisposable(() {
-      if (!controller.hasListener) {
+      if (!controller.hasListener && !controller.isClosed && !isDone) {
         controller.stream.listen((_) {});
       }
       return controller.close();
```
bencampbell-wf: Nothing jumps out at me as a possible leak.

I would suggest, possibly in another PR, to implement an internal implementation of `manageDisposable` in which `_InternalDisposable`s are removed from the list `_internalDisposables` when disposed. This will provide an added defense against leaking by removing anything bound by the `_InternalDisposable` from the class scope and allowing GC to do its thing.

I can't think of any reason why the implementation would need to hold onto these `_InternalDisposable` after they have been disposed.

test/unit/vm/disposable_test.dart
```diff
@@ -262,8 +262,29 @@ void main() {
       });
 
       test(
-          'should close a single-subscription stream with no listener'
-          'when parent is disposed', () async {
+          'should complete normally for a single-subscription stream with '
+          'a listener that has been closed when parent is disposed', () async {
```
bencampbell-wf: 🌵 This description reads a little funky.

```diff
-          'should complete normally for a single-subscription stream with '
-          'a listener that has been closed when parent is disposed', () async {
+          'should allow subscriptions for a single-subscription stream to be '
+          'cancelled after parent has been disposed', () async {

**Comment by trentgrover-wf** [comment]
+1

**Comment by bencampbell-wf** [comment]
+1

File: lib/src/disposable/disposable.dart [hunk 3]
```diff
@@ -300,8 +300,20 @@ class Disposable implements _Disposable, DisposableManagerV3 {
   @override
   void manageStreamController(StreamController controller) {
     _throwOnInvalidCall('manageStreamController', 'controller', controller);
+    // If a single-subscription stream has a subscription and that
+    // subscription is subsequently canceled, the `done` future will
+    // complete, but there is no other way for us to tell that this
+    // is what has happened. If we then listen to the stream (since
+    // closing a stream that was never listen-to throws) we'll get
```
evanweible-wf: I was under the impression that closing a single-subscription StreamController that had not been listened to would just never complete, not throw

evanweible-wf: also I think `listen-to` should be `listened to`

georgelesica-wf: Hmm, yeah, I think you're right. I thought it was throwing because when I wrote the tests (TDD FTW) it threw. But there were two actual problems and I was probably only hitting one of them in both tests.

lib/src/disposable/disposable.dart
```diff
@@ -300,12 +300,26 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
+    var disposable = new _InternalDisposable(() {
+      if (!controller.hasListener && !controller.isClosed && !isDone) {
         controller.stream.listen((_) {});
       }
       return controller.close();
-    }));
+    });
+    controller.done.then((_) {
+      isDone = true;
+      disposable.dispose();
```
bencampbell-wf: Did you want to remove the `disposable` instance from the `_internalDisposables` list on done?

lib/src/disposable/disposable.dart
```diff
@@ -30,10 +30,10 @@ class _InternalDisposable implements _Disposable {
 
   @override
   Future<Null> dispose() {
-    var disposeFuture = _disposer();
+    var disposeFuture = _disposer != null ? _disposer() : null;
```
bencampbell-wf: 🔮 Why not use:

```dart
var disposeFuture = _disposer?.call();
```

georgelesica-wf: Yeah that'd work too.

seanburke-wf:
+10
CI passes. Also tested against the evidence testing experience in graph_app, where we were seeing the exception that motivated the bug report. The exception in question is gone. Got the following error immediately before reload during testing, but was able to reproduce against the 1.4.0 release:

```
VM999:1 Uncaught Unhandled exception:
NoSuchMethodError: The getter '_value' was called on null.
Receiver: null
Tried calling: _value
#0      Object._noSuchMethod (dart:core-patch/object_patch.dart:44)
#1      Object.noSuchMethod (dart:core-patch/object_patch.dart:47)
#2      DateTime.difference (dart:core-patch/date_patch.dart:179)
#3      InteractionObserver._handleThrottledInteraction (package:app_intelligence/src/app_data_services/browser/interaction_observer.dart:173:35)
#4      _RootZone.runUnaryGuarded (dart:async/zone.dart:1338)
#5      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:331)
#6      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:258)
#7      _SyncBroadcastStreamController._sendData (dart:async/broadcast_stream_controller.dart:386)
#8      _BroadcastStreamController.add (dart:async/broadcast_stream_controller.dart:261)
#9      _EventStreamSubscription._EventStreamSubscription.<anonymous closure> (dart:html:44616)(anonymous function) @ VM999:1
```

todbachman-wf:
+1

maxwellpeterson-wf:
+1

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - Note: final commit does not change any of the code
  - N/A
- [x] All unit tests pass

Merging into master

PR #32, opened 2017-06-08
RAP-2040 Large number of managed disposables cause massive GC hit

georgelesica-wf:
> - bug / critical
> - feature

### Description

When there are a large number of managed disposables (10k+), disposal of the managing object causes a large number of garbage collections apparently due to the list of futures that we build in the dispose() method.

### Changes

This isn't a perfect solution (yet) but I think it improves things a bit. We don't build a list of futures. Instead we await them one at a time in a for-loop.

This would perform much better if we didn't have to check to `isDisposedOrDisposing` in the callbacks we set up on management. However, these are necessary, for now, because the documented semantics allow most managed objects to be disposed / closed out-of-band. This means we have to handle that event to prevent memory leaks, so on dispose we have to remove each one from the list of disposables we maintain.

### Semantic Versioning

- [x] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes
- [ ] See me for a +10 walkthrough

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/disposable/disposable.dart
+++ b/lib/src/disposable/disposable.dart
@@ -14,6 +14,7 @@
 
 import 'dart:async';
 
+import 'dart:collection';
 import 'package:meta/meta.dart';
 
 import 'package:w_common/src/disposable/disposable_manager.dart';
@@ -165,9 +166,9 @@ typedef Future<dynamic> Disposer();
 /// composition to include the [Disposable] machinery without changing
 /// the public interface of our class or polluting its lifecycle.
 class Disposable implements _Disposable, DisposableManagerV3 {
-  List<Future<dynamic>> _awaitableFutures = <Future<dynamic>>[];
+  final Set<Future> _awaitableFutures = new HashSet<Future>();
   Completer<Null> _didDispose = new Completer<Null>();
-  List<_Disposable> _internalDisposables = <_Disposable>[];
+  final Set<_Disposable> _internalDisposables = new HashSet<_Disposable>();
   bool _isDisposing = false;
 
   /// A [Future] that will complete when this object has been disposed.
@@ -196,9 +197,13 @@ class Disposable implements _Disposable, DisposableManagerV3 {
     _throwOnInvalidCall('awaitBeforeDispose', 'future', future);
     _awaitableFutures.add(future);
     future.then((_) {
-      _awaitableFutures.remove(future);
+      if (!isDisposedOrDisposing) {
+        _awaitableFutures.remove(future);
+      }
     }).catchError((_) {
-      _awaitableFutures.remove(future);
+      if (!isDisposedOrDisposing) {
+        _awaitableFutures.remove(future);
+      }
     });
     return future;
   }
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/disposable/disposable.dart
```diff
@@ -196,9 +197,13 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
     future.then((_) {
-      _awaitableFutures.remove(future);
+      if (!isDisposedOrDisposing) {
+        _awaitableFutures.remove(future);
+      }
     }).catchError((_) {
-      _awaitableFutures.remove(future);
+      if (!isDisposedOrDisposing) {
+        _awaitableFutures.remove(future);
+      }
```
bencampbell-wf: 🔮  You could use `whenComplete` to handle the error and success case in one statement.

lib/src/disposable/disposable.dart
```diff
@@ -273,9 +280,13 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
     completer.future.catchError((e) {
-      _internalDisposables.remove(disposable);
+      if (!isDisposedOrDisposing) {
+        _internalDisposables.remove(disposable);
+      }
     }).then((_) {
-      _internalDisposables.remove(disposable);
+      if (!isDisposedOrDisposing) {
+        _internalDisposables.remove(disposable);
+      }
```
bencampbell-wf: 🔮  You could use `whenComplete` to handle the error and success case in one statement.

lib/src/disposable/disposable.dart
```diff
@@ -214,19 +219,17 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
-    List<Future<dynamic>> futures = []
-      ..addAll(_internalDisposables.map((disposable) => disposable.dispose()))
-      ..add(onDispose());
-    _internalDisposables = [];
+    await Future.wait(_awaitableFutures);
+    _awaitableFutures.clear();
+
+    for (var disposable in _internalDisposables) {
+      await disposable.dispose();
+    }
```
bencampbell-wf: :man_shrugging: I'd be interested in seeing the change in the JS output to see what is changing under the hood.

georgelesica-wf: Yeah, that's what we thought. Once we get this squared away I'm going create a minimal test case and try to get to the bottom of it.

lib/src/disposable/disposable.dart
georgelesica-wf: Actually, I tried that just now and noticed that `whenComplete` re-raises exceptions, which we don't want in this case. Maybe there's a workaround.

todbachman-wf:
+10

#### Testing Performed:
- [x] Confirm CI passes
- [x] Link into docs
- [x] In Chrome, open a document with XBRL enabled
- [x] Profile the behavior of closing a document
- [x] Confirm the time to close the document is improved (previously 20 seconds to close)
- [x] Confirm memory consumption when the document is closed is improved

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - N/A
  - N/A
- [x] All unit tests pass
- [ ] Release notes provided

Merging into master

PR #35, opened 2017-06-22
RAP-2067 Add event listener helpers to Disposable.

smaifullerton-wk:
### Description
Currently when consumers call `addEventListener`, they also have to manually `removeEventListener` before disposal, or the listeners stay in memory. 

It would be better if consumers could use helper methods on `Disposable` that will automatically keep track and manage the removal of the listeners upon disposal.

### Changes
I added helper methods for `window`, `document`, and `element`s to automatically handle the `removeEventListener` call upon disposal. 

This PR was a great learning opportunity for dependency injection vs. integration testing, so that is why I tested with an integration test for the `element` method and used a mock class for the `window` and `document` methods so as to not mutate any global states while testing.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf


aviary2-wf:
## Raven
Number of Findings: 0

dustinlessard-wf:
+1

lib/src/disposable/disposable.dart
```diff
@@ -427,6 +428,28 @@ class Disposable implements _Disposable, DisposableManagerV3 {
     return null;
   }
 
+  void subscribeToDocumentEvent(String event, EventListener callback,
```
bencampbell-wf: Are these methods we would want to expose on the `DisposableManager` interface? These methods would not be available to `LifecycleModule` without doing so.

smaifullerton-wk: It looks like all the managing methods on `Disposable` are exposed that way, so probably.

We have a separate ticket for adding a `subscribeToStream` method to `Disposable`, so maybe we should wait until that's done and add all of these new methods to `DisposableManager` at once, so we only have to deprecate once? @georgelesica-wf what do you think?

georgelesica-wf: Yeah, we should add them to those interfaces, we'll need to add a new `DisposableManagerV4` interface as well so we don't break consumers. But I agree with @smaifullerton-wk that maybe we should wait to do that in RAP-2068 so that we don't end up with `DisposableManagerV5 like 48 hours after v4.

lib/src/disposable/disposable.dart
```diff
@@ -14,6 +14,7 @@
 
 import 'dart:async';
 import 'dart:collection';
+import 'dart:html';
```
georgelesica-wf: Tomorrow let's look at how we might be able to remove this dependency, or at least mitigate it.

lib/src/disposable/disposable.dart
```diff
@@ -427,6 +428,28 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
+  }
+
+  void subscribeToDomElementEvent(
+      Element element, String event, EventListener callback,
+      {bool useCapture: null}) {
+    _subscribeToEvent(element, event, callback, useCapture);
+  }
+
+  void subscribeToWindowEvent(String event, EventListener callback,
+      {bool useCapture: null, EventTarget windowObject: null}) {
```
georgelesica-wf: Default value is already null, so you can leave those off. You might say the "default default value" is null...

georgelesica-wf:
Let's also think about whether we can tidy up after ourselves if the consumer happens to call `removeEventListener` themselves. Ideally we don't want them to do that, but it would be cool if we could support it for convenience.

dustinlessard-wf:
Are we at a point with some of this stuff, where we could benefit consumers with more thorough documentation, like best-practices and common uses?

lib/src/disposable/disposable.dart
bencampbell-wf: However you'd like to coordinate it is fine.

lib/src/disposable/disposable.dart
bencampbell-wf: I don't know if I'd support a `null` default value for a boolean. It seems like the strategy is to defer the default value to whatever `eventTarget.addEventListener` decides. If a default value is undecided, I would suggest a doc comment explaining how a default is arrived and why you may want to choose something other than the default.

lib/src/disposable/disposable.dart
bencampbell-wf: 👼 I would make a note to include doc comments when the `DisposableManager` interface is updated.

lib/src/disposable/disposable.dart
smaifullerton-wk: Per us looking at this, we've decided that there isn't enough of a need right now to export both `disposable_browser` and `disposable_vm` to allow consumers to use this class without methods that require the html library. Down the line we could make this change if necessary.

smaifullerton-wk:
`EventTarget` doesn't provide a way for us to find out if its listener has been removed, so we can't allow for manual usage of the `removeEventListener` method. I added doc comments to the new methods to clarify.

georgelesica-wf:
@dustinlessard-wf Let's do some documentation expansion in a separate ticket, I'll cut one... RAP-2103.

test/unit/browser/disposable_test.dart
```diff
@@ -350,5 +352,69 @@ void main() {
...
+
+      test(
+          'subscribeToDomElementEvent should remove listener when thing is disposed',
+          () async {
+        var element = new Element.div();
+        int numberOfEventCallbacks = 0;
+        EventListener callback = (_) {
+          numberOfEventCallbacks++;
+        };
+        var event = new Event('event');
```
georgelesica-wf: It might be nice to extract the string "event" into a variable like `eventName` to be consistent with the previous test and the event objects in this test. #nit

test/unit/browser/disposable_test.dart
```diff
@@ -350,5 +352,69 @@ void main() {
...
+        var shouldNotListenEvent = new Event('shouldNotListenEvent');
+
+        thing.subscribeToDomElementEvent(element, 'event', callback);
+        expect(numberOfEventCallbacks, equals(0));
+
+        element.dispatchEvent(shouldNotListenEvent);
```
georgelesica-wf: Could you assert that it's zero here, otherwise the wrong one might have caused it to increment (not likely, I realize).

test/unit/browser/disposable_test.dart
```diff
@@ -350,5 +352,69 @@ void main() {
...
+        await thing.dispose();
+
+        element.dispatchEvent(event);
+        expect(numberOfEventCallbacks, equals(1));
+
+        thing.subscribeToDomElementEvent(element, 'event', callback);
+        element.dispatchEvent(event);
+        element.dispatchEvent(event);
+        element.dispatchEvent(shouldNotListenEvent);
+        expect(numberOfEventCallbacks, equals(3));
```
georgelesica-wf: Same here, I would make multiple assertions to make sure the right events are actually causing it to increment.

test/unit/browser/disposable_test.dart
```diff
@@ -350,5 +352,69 @@ void main() {
...
+        var window = new MockEventTarget();
+        var event = 'event';
+        EventListener callback = (_) {};
+        var useCapture = true;
+
+        thing.subscribeToWindowEvent(event, callback,
+            windowObject: window, useCapture: useCapture);
+        verify(window.addEventListener(event, callback, useCapture));
+        await thing.dispose();
+        verify(window.removeEventListener(event, callback, useCapture));
```
georgelesica-wf: This test and the first test could probably share some of their setup if you put them in a group with a `setUp` callback. #nit

georgelesica-wf:
Nice work, just a couple small comments on the tests!

todbachman-wf:
+10. Confirmed CI passed.

todbachman-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
  - N/A
- [x] All unit tests pass

Merging into master

PR #42, opened 2017-06-28
RAP-2123 Enable mixing in browser version of Disposable

todbachman-wf:
### Description
The browser version of `Disposable` can't be used as a mixin anymore because it extends the VM version. We've been encouraging users to use it as a mixin, and they have been, so this is an issue.

### Changes
Updated the browser version of `Disposable` to implement the VM version instead of extending it. It now proxies to an internal instance of the VM disposable.

### Semantic Versioning
- [x] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA
- [ ] CI passes

### Code Review
@Workiva/rich-app-platform-pp 
@Workiva/web-platform-pp


aviary-wf:
## Raven
Number of Findings: 0

georgelesica-wf:
+1

lib/src/browser/disposable_browser.dart
```diff
@@ -12,11 +12,191 @@
...
+class _InnerDisposable extends disposable_common.Disposable {
+  Func<Future<Null>> onDisposeHandler;
+
+  @override
+  Future<Null> onDispose() {
+    return onDisposeHandler();
+  }
+}
+
+/// Allows the creation of managed objects, including helpers for common
```
todbachman-wf: I copy/pasted the doc comment from the VM version of `Disposable` here since I don't think it would otherwise be picked up by the API docs.

lib/src/browser/disposable_browser.dart
```diff
@@ -12,11 +12,191 @@
...
+import 'dart:async';
 import 'dart:html';
 
+import 'package:meta/meta.dart';
+import 'package:w_common/func.dart';
+
 import 'package:w_common/src/common/disposable.dart' as disposable_common;
 
-class Disposable extends disposable_common.Disposable {
+class _InnerDisposable extends disposable_common.Disposable {
```
todbachman-wf: This is the class that will be instantiated internally in the browser version of Disposable and all calls will be proxied to it. It allows us to call `Disposable.onDispose` when `dispose()` is called.

lib/src/browser/disposable_browser.dart
```diff
@@ -60,11 +240,8 @@ class Disposable extends disposable_common.Disposable {
   void _subscribeToEvent(EventTarget eventTarget, String event,
       EventListener callback, bool useCapture) {
     eventTarget.addEventListener(event, callback, useCapture);
-
-    var disposable = new disposable_common.InternalDisposable(() {
+    _disposable.manageDisposer(() {
       eventTarget.removeEventListener(event, callback, useCapture);
     });
-
-    disposable_common.addInternalDisposable(this, disposable);
```
todbachman-wf: I changed the logic to create and manage a `Disposer` so we no longer need to jump through the hoops provided by `addInternalDisposable`.

test/unit/disposable_common.dart
```diff
@@ -0,0 +1,347 @@
+import 'dart:async';
+
+import 'package:test/test.dart';
+import 'package:w_common/disposable.dart';
+import 'package:w_common/func.dart';
+
+import 'stubs.dart';
+
+void testCommonDisposable(Func<StubDisposable> disposableFactory) {
```
todbachman-wf: I added this utility method to allow us to test the common behavior on both the VM- and browser-based versions of `Disposable`.

aaronstgeorge-wf:
+10   

Tested
-------

- [ ] Ci

aaronstgeorge-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass
- [x] **Dependency Scan Clean**

Merging into master.

PR #44, opened 2017-06-28
RAP-2125 Add leak flag functionality

georgelesica-wf:
### Description

@greglittlefield-wf suggested that if we annotated disposable objects with instances of a special class upon disposal, we would be better able to find things that were being leaked by searching for the special class in a heap snapshot. In theory, any instance that shows up is a leak (assuming the GC has run).

Here's Greg's example https://gist.github.com/greglittlefield-wf/4703ff40c98800047e279ca73df715b7

### Changes

Add a `LeakFlag` and corresponding attribute to `Disposable`.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -148,9 +148,6 @@ class Disposable implements disposable_common.Disposable {
   bool get isDisposing => _disposable.isDisposing;
 
   @override
-  String get leakFlagDescription => _disposable.leakFlagDescription;
-
-  @override
   Future<T> awaitBeforeDispose<T>(Future<T> future) => _disposable
       .awaitBeforeDispose(future);
 
@@ -161,7 +158,7 @@ class Disposable implements disposable_common.Disposable {
   }
 
   @override
-  void flagLeak(String description) {
+  void flagLeak([String description]) {
     _disposable.flagLeak(description);
   }
 
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -148,6 +148,9 @@ class Disposable implements disposable_common.Disposable {
   bool get isDisposing => _disposable.isDisposing;
 
   @override
+  String get leakFlagDescription => _disposable.leakFlagDescription;
+
+  @override
   Future<T> awaitBeforeDispose<T>(Future<T> future) => _disposable
       .awaitBeforeDispose(future);
 
@@ -155,6 +158,11 @@ class Disposable implements disposable_common.Disposable {
   Future<Null> dispose() {
     _disposable.onDisposeHandler = this.onDispose;
     return _disposable.dispose();
+  }
+
+  @override
+  void flagLeak(String description) {
+    _disposable.flagLeak(description);
   }
 
   @override
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -148,6 +148,9 @@ class Disposable implements disposable_common.Disposable {
   bool get isDisposing => _disposable.isDisposing;
 
   @override
+  bool get isLeakFlagSet => _disposable.isLeakFlagSet;
+
+  @override
   Future<T> awaitBeforeDispose<T>(Future<T> future) => _disposable
       .awaitBeforeDispose(future);
 
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -157,7 +157,12 @@ class Disposable implements disposable_common.Disposable {
   @override
   Future<Null> dispose() {
     _disposable.onDisposeHandler = this.onDispose;
-    return _disposable.dispose();
+    return _disposable.dispose().then((_) {
+      // We want the description to be the runtime type of this
+      // object, not the proxy disposable, so we need to reset
+      // the leak flag here.
+      flagLeak(runtimeType.toString());
+    });
   }
 
   @override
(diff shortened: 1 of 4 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/common/disposable.dart
```diff
@@ -284,6 +300,14 @@ class Disposable implements _Disposable, DisposableManagerV3 {
...
           '$runtimeType $hashCode took ${stopwatch.elapsedMicroseconds / 1000000.0} seconds to dispose');
     }
+
+    flagLeak(runtimeType.toString());
+  }
+
+  @mustCallSuper
+  @override
+  void flagLeak(String description) {
+    _leakFlag = new LeakFlag(description);
```
evanweible-wf: just to verify, this wouldn't _need_ to be overridden by consumers of Disposable, right?

greglittlefield-wf: Yeah, I think ideally, this should never get overridden.

lib/src/common/disposable.dart
```diff
@@ -284,6 +300,14 @@ class Disposable implements _Disposable, DisposableManagerV3 {
       _logger.info(
           '$runtimeType $hashCode took ${stopwatch.elapsedMicroseconds / 1000000.0} seconds to dispose');
     }
+
+    flagLeak(runtimeType.toString());
```
bencampbell-wf: Would you wan to pass `leakFlagDescription` here instead of `runtimeType`?

bencampbell-wf: Derp, never mind. I assumed leakFlag was meant to provide consumers a way to add a more explicit description that would be used to describe the leak flag.

lib/src/common/disposable.dart
todbachman-wf: Do we expect consumers to decorate other classes with the `LeakFlagger` mixin, or do we only expect it to be used in `Disposable`? If we only expect `Disposable` to use it, I'd be inclined to not bother with the `LeakFlagger` mixin at all and then just make this a private method.

lib/src/common/disposable.dart
```diff
@@ -215,6 +228,9 @@ class Disposable implements _Disposable, DisposableManagerV3 {
     return size;
   }
 
+  /// The description of the currently-set leak flag.
+  String get leakFlagDescription => _leakFlag.description;
```
todbachman-wf: How would this be used?

todbachman-wf: It seems like having it on the `_leakFlag` itself should be good enough.

georgelesica-wf:
@todbachman-wf I was thinking consumers like `LifecycleModule` that use the interfaces might want to implement `LeakFlagger` as well. Alternatively I could integrate it with `DisposableManager` and make it part of V4 so every who uses the manager interface picks it up and, presumably, pushes it through to their proxy.

todbachman-wf:
@georgelesica-wf I forgot that at least `LifecycleModule` would also need to implement this. I'm cool with leaving `LeakFlagger` as a separate mixin then.

lib/src/browser/disposable_browser.dart
```diff
@@ -158,6 +161,11 @@ class Disposable implements disposable_common.Disposable {
   }
 
   @override
+  void flagLeak([String description]) {
```
todbachman-wf: since we proxy to an internal disposable object, this method will never be called by `_disposable.dispose()`. Does it make sense to provide some sort of hook in the `_InnerDisposable` class like we did for the `onDispose` method? If we don't do something the parent class of the `LeakFlag` in the heap trace will always be `_InnerDisposable`.

todbachman-wf: What I should have said is that the parent is always going to be `_InnerDisposable`, but we could use the description to provide better context to consumers.

lib/src/browser/disposable_browser.dart
```diff
@@ -148,13 +148,26 @@ class Disposable implements disposable_common.Disposable {
...
+      // We want the description to be the runtime type of this
+      // object, not the proxy disposable, so we need to reset
+      // the leak flag here.
+      flagLeak(runtimeType.toString());
+    });
+  }
+
+  @override
+  void flagLeak([String description]) {
+    _disposable.flagLeak(description);
```
todbachman-wf: Probably want to add the debug mode check here too.

lib/src/common/disposable.dart
```diff
@@ -74,6 +74,17 @@ class _ObservableTimer implements Timer {
...
 }
 
+/// A class used as a marker for potential memory leaks.
+class LeakFlag {
+  final String description;
+
+  LeakFlag(this.description);
+
+  @override
+  String toString() =>
```
jayudey-wf: would it be worth adding a test for htis?

georgelesica-wf: It seems trivially correct to me and there's no reason it has to be anything in particular since we don't really use it anywhere.

jayudey-wf: ok

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI exercising new functionality
- [ ] All unit tests pass

Merging into master.

PR #45, opened 2017-07-12
RAP-2147 Fix timer factory bug

georgelesica-wf:
### Description

We allow the managed timer factories to be called during / after disposal but this will often (or perhaps always) result in a rather obscure exception because we attempt to add to the `_internalDisposables` list while it is being iterated.

A quick fix would have been to simply copy the list before iterating, but this could have actually resulted in memory leaks, particularly in the case of periodic timers since a new timer would have been created and then never forcibly disposed.

A better fix is to simply throw if one of those factories is called during / after disposal. This is the pattern we follow for the other management methods.

### Changes

Throw if one of the managed timer factories is invoked during or after disposal.

### Semantic Versioning

- [x] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/test/unit/disposable_common.dart
+++ b/test/unit/disposable_common.dart
@@ -60,6 +60,29 @@ void testCommonDisposable(Func<StubDisposable> disposableFactory) {
       });
       disposable.dispose();
     });
+
+    test('should throw during disposal', () {
+      var completer = new Completer();
+      disposable.awaitBeforeDispose(completer.future);
+      disposable.dispose();
+      new Future(() {}).then((_) {
+        expect(disposable.isDisposing, isTrue);
+        expect(
+            () => disposable.getManagedDelayedFuture(
+                new Duration(seconds: 10), () => null),
+            throwsStateError);
+        completer.complete();
+      });
+    });
+
+    test('should throw after disposal', () async {
+      await disposable.dispose();
+      expect(disposable.isDisposed, isTrue);
+      expect(
+          () => disposable.getManagedDelayedFuture(
+              new Duration(seconds: 10), () => null),
+          throwsStateError);
+    });
   });
 
   group('getManagedTimer', () {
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
<details>

</details>

test/unit/disposable_common.dart
```diff
@@ -60,6 +60,29 @@ void testCommonDisposable(Func<StubDisposable> disposableFactory) {
...
+      var completer = new Completer();
+      disposable.awaitBeforeDispose(completer.future);
+      disposable.dispose();
+      new Future(() {}).then((_) {
+        expect(disposable.isDisposing, isTrue);
+        expect(
+            () => disposable.getManagedDelayedFuture(
+                new Duration(seconds: 10), () => null),
+            throwsStateError);
+        completer.complete();
```
bencampbell-wf: 🌵 You could probably DRY these tests up.

test/unit/disposable_common.dart
```diff
@@ -60,6 +60,29 @@ void testCommonDisposable(Func<StubDisposable> disposableFactory) {
       });
       disposable.dispose();
     });
+
+    test('should throw during disposal', () {
+      var completer = new Completer();
+      disposable.awaitBeforeDispose(completer.future);
+      disposable.dispose();
+      new Future(() {}).then((_) {
```
evanweible-wf: need to await this or use `expectAsync()` otherwise the `then()` callback can/will execute after the test is finished

evanweible-wf: same comment for the other two tests like this one

georgelesica-wf: That's what I thought, but I tested it and it seemed to work. Is it intermittent?

evanweible-wf: It might have just been completing fast enough, not sure

jayudey-wf:
+10, CI did indeed pass

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
- [ ] Dev/QA +10 with detail of what was tested (see above)
- [ ] All unit tests pass

Merging into master.

PR #46, opened 2017-07-13
RAP-2068 managed stream subscriptions

aaronstgeorge-wf:
### Description

Disposable doesn't know when `StreamsSubscription`s are cancelled. If a consumer manually cancels a `StreamSubscription` disposable will keep an internal reference around until explicit disposal of the parent object.

### Changes

Create `ManagedStreamSubscription` which has a `onClose` future. This anchor point allows `Disposable` to know when streams are cancelled. It can then remove references and allow the subscription to be garbage collected.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf


lib/src/common/disposable.dart
```diff
@@ -19,6 +19,7 @@ import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
 import 'package:w_common/src/common/disposable_manager.dart';
+import 'package:w_common/src/common/managed_stream_subscrption.dart';
```
smaifullerton-wk: `managed_stream_subscrption` -> `managed_stream_subscription`

aviary2-wf:
## Raven
<details>

</details>

lib/src/common/disposable.dart
```diff
@@ -351,6 +352,32 @@ class Disposable implements _Disposable, DisposableManagerV3, LeakFlagger {
 
   @mustCallSuper
   @override
+  StreamSubscription<T> getManagedStreamSubscription<T>(
+      Stream<T> stream, void onData(T event)) {
+    _throwOnInvalidCall2('awaitBeforeDispose', 'future', stream, onData);
```
aaronstgeorge-wf: I tried to follow the same naming convention as `expectAsync` `expectAsync2`, ect.

lib/src/common/disposable_manager.dart
```diff
@@ -112,6 +134,25 @@ abstract class DisposableManagerV3 implements DisposableManagerV2 {
...
+/// This interface allows consumers to exercise more control over how
+/// disposal is implemented for their classes.
+///
+/// When new management methods are to be added, they should be added
+/// here first, then implemented in [Disposable].
+abstract class DisposableManagerV4 implements DisposableManagerV3 {
+  /// Automatically cancel a stream subscription when this object is disposed.
+  ///
+  /// If the returned stream subscription is managed manually (i.e. canceled
+  /// before disposal of the parent object) [Disposable] will properly manage the
```
aaronstgeorge-wf: `the` -> `its`

lib/src/common/disposable.dart
georgelesica-wf: I've just made two calls to `_throwOnInvalidCall` in the past. There is a tiny bit of repetition, but not much. This is fine, though.

lib/src/common/disposable.dart
```diff
@@ -436,6 +463,7 @@ class Disposable implements _Disposable, DisposableManagerV3, LeakFlagger {
     _internalDisposables.add(disposable);
   }
 
+  @deprecated
```
georgelesica-wf: We should discuss whether we actually want to deprecate this or not. Can anyone think of any edge cases where the new method won't work? @Workiva/rich-app-platform-pp

lib/src/common/disposable.dart
```diff
@@ -505,4 +533,12 @@ class Disposable implements _Disposable, DisposableManagerV3, LeakFlagger {
           '$methodName not allowed, object is already disposed');
     }
   }
+
+  void _throwOnInvalidCall2(String methodName, String parameterName,
+      dynamic parameterValue, dynamic secondParameterValue) {
```
georgelesica-wf: What about the second parameter name?

lib/src/common/disposable_manager.dart
```diff
@@ -112,6 +134,25 @@ abstract class DisposableManagerV3 implements DisposableManagerV2 {
...
+///
+/// This interface allows consumers to exercise more control over how
+/// disposal is implemented for their classes.
+///
+/// When new management methods are to be added, they should be added
+/// here first, then implemented in [Disposable].
+abstract class DisposableManagerV4 implements DisposableManagerV3 {
+  /// Automatically cancel a stream subscription when this object is disposed.
+  ///
+  /// If the returned stream subscription is managed manually (i.e. canceled
```
georgelesica-wf: I would just say "canceled manually".

lib/src/common/disposable.dart
aaronstgeorge-wf: Ooops 🤕

lib/src/common/disposable_manager.dart
georgelesica-wf: s/properly manage/clean up/

lib/src/common/disposable_manager.dart
```diff
@@ -112,6 +134,25 @@ abstract class DisposableManagerV3 implements DisposableManagerV2 {
...
+  /// before disposal of the parent object) [Disposable] will properly manage the
+  /// internal reference.
+  ///
+  /// Neither parameter may not be `null`.
```
georgelesica-wf: So they both *must* be null? :-)

lib/src/common/disposable_manager.dart
```diff
@@ -112,6 +134,25 @@ abstract class DisposableManagerV3 implements DisposableManagerV2 {
...
+  StreamSubscription<T> getManagedStreamSubscription<T>(
```
georgelesica-wf: Let's talk about the name. In this case the goal is really not so much to "get" a new subscription, but to subscribe to a stream. The subscription is really just a side effect. Can anyone think of a better name, given this?

lib/src/common/managed_stream_subscription.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:async';
+
+/// An implementation of StreamSubscription that provides a didCancel future.
```
georgelesica-wf: Can we flesh this out a bit more? Why does that matter? How would it be used?

lib/src/common/managed_stream_subscription.dart
```diff
@@ -0,0 +1,42 @@
...
+  Completer<Null> _didCancel = new Completer();
+
+  ManagedStreamSubscription(Stream<T> stream, void onData(T))
+      : _subscription = stream.listen(onData);
+
+  @override
+  Future<E> asFuture<E>([E futureValue]) => _subscription.asFuture(futureValue);
+
+  @override
+  Future<Null> cancel() async {
```
georgelesica-wf: Let's not make this async since it isn't on the base class. Maybe just chain the completion onto the end of the subscription's cancel future. Although that can be null, so you'll want to null check it and build a throwaway future (maybe) otherwise.

lib/src/common/managed_stream_subscription.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:async';
+
+/// An implementation of StreamSubscription that provides a didCancel future.
+class ManagedStreamSubscription<T> extends StreamSubscription<T> {
```
georgelesica-wf: Why not `implements` instead of `extends`?

aaronstgeorge-wf: No reason. The implementation isn't publicly exported so the two are equivalent.

georgelesica-wf: If `StreamSubscription` were to change out from under us, though, we'd end up with weird bugs instead of static errors.

lib/src/common/disposable_manager.dart
aaronstgeorge-wf: I chose this to be consistent with `getManagedTimer`, and `getManagedPeriodicTimer`. But I agree with your point. We could just do `subscribeToStream` but I would worry that the name doesn't convey that the returned stream subscription is managed in a different way than `manageStreamSubscription` previously was.

georgelesica-wf: Yeah, I'm not entirely sure how I feel. I could go either way. If we're going to rename it, however, I prefer `listenToStream` since it stays consistent with the `listen` method on streams.

lib/src/browser/disposable_browser.dart
```diff
@@ -184,6 +184,11 @@ class Disposable implements disposable_common.Disposable {
       _disposable.getManagedTimer(duration, callback);
 
   @override
+  StreamSubscription<T> getManagedStreamSubscription<T>(
+          Stream<T> stream, void onData(T event)) =>
```
bencampbell-wf: 🔥 You need optional params for `onError`, `onDone`, etc... 

I wonder if the Dart SDK has a typedef for the listen method on streams.

lib/src/common/disposable_manager.dart
```diff
@@ -112,6 +134,25 @@ abstract class DisposableManagerV3 implements DisposableManagerV2 {
...
+abstract class DisposableManagerV4 implements DisposableManagerV3 {
+  /// Automatically cancel a stream subscription when this object is disposed.
+  ///
+  /// If the returned `StreamSubscription` is cancelled manually (i.e. canceled
+  /// before disposal of the parent object) [Disposable] will clean up the
+  /// internal reference allowing the subscription to be garbage collected.
+  ///
+  /// Neither parameter may be `null`.
+  StreamSubscription<T> getManagedStreamSubscription<T>(
+      Stream<T> stream, void onData(T event));
```
bencampbell-wf: 🔥 You need optional params for `onError`, `onDone`, etc... 

I wonder if the Dart SDK has a typedef for the listen method on streams.

lib/src/common/managed_stream_subscription.dart
```diff
@@ -0,0 +1,49 @@
...
+    });
+  }
+
+  Future<Null> get didCancel => _didCancel.future;
+
+  @override
+  bool get isPaused => _subscription.isPaused;
+
+  @override
+  void onData(void handleData(_)) => _subscription.onData(handleData);
```
bencampbell-wf: 🌵 You should use the type parameter here.

```diff
-  void onData(void handleData(_)) => _subscription.onData(handleData);
+  void onData(void handleData(T _)) => _subscription.onData(handleData);
```

lib/src/common/managed_stream_subscription.dart
```diff
@@ -0,0 +1,49 @@
...
+            onError: onError, onDone: onDone, cancelOnError: cancelOnError);
+
+  @override
+  Future<E> asFuture<E>([E futureValue]) => _subscription.asFuture(futureValue);
+
+  @override
+  Future<Null> cancel() {
+    return (_subscription.cancel() ?? new Future(() {})).then((_) {
+      if (!_didCancel.isCompleted) {
+        _didCancel.complete(null);
```
bencampbell-wf: 🔮 You can omit the `null` value here:

```diff
-        _didCancel.complete(null);
+        _didCancel.complete();

lib/src/common/disposable.dart
bencampbell-wf: Are there any instances where consumers are passing stream subscription references between classes?

I know there are some instances where stream subscription lifetimes are less that that of their disposable parent. Consumers often use lists of subscriptions to batch them for cancellation. I would be interested if that behavior would extend across class instances.

georgelesica-wf:
+1

georgelesica-wf:
+10 CI passed

georgelesica-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
- [x] Unit tests created/updated 
- [x] All unit tests pass
- [ ] Rosie has run and reports properly the release the ticket will be included in - NA (OSS)
- [x] Security review completed if necessary

Merging into master.
@Workiva/release-management-pp

rmconsole-wf:
+1 from RM

PR #47, opened 2017-07-20
RAP-2157 Return ManagedDisposer from getManagedDisposer.

aaronstgeorge-wf:
[Jira Ticket](https://jira.atl.workiva.net/browse/RAP-2157)

Problem
--------

Currently there is no way for `Disposable` to know if a `Disposer` has been short circuited, e.g. called before disposal of the object. References held in redundant disposers have been a vector for leaks in the past.

Solution
--------

If `manageDisposer` instead returned a `Disposable` that could be called by consumers when they intended the disposers functionality to be invoked, `Disposable` would know when a `Disposer` has become redundant and could therefore be removed from internal collections. To this end `SimpleDisposable` was created and a new method `getManagedDisposer`. `manageDisposer` was deprecated.

Tests
--------
Added tests for new functionality.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf
@todbachman-wf 
@bencampbell-wf 
@georgelesica-wf

### DIFF
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -19,6 +19,7 @@ import 'package:meta/meta.dart';
 import 'package:w_common/func.dart';
 
 import 'package:w_common/src/common/disposable.dart' as disposable_common;
+import 'package:w_common/src/common/disposable_manager.dart';
 
 class _InnerDisposable extends disposable_common.Disposable {
   Func<Future<Null>> onDisposeHandler;
@@ -58,7 +59,7 @@ class _InnerDisposable extends disposable_common.Disposable {
 ///        }
 ///      }
 ///
-/// The [manageDisposer] helper allows you to clean up arbitrary objects
+/// The [getManagedDisposer] helper allows you to clean up arbitrary objects
 /// on dispose so that you can avoid keeping track of them yourself. To
 /// use it, simply provide a callback that returns a [Future] of any
 /// kind. For example:
@@ -68,14 +69,14 @@ class _InnerDisposable extends disposable_common.Disposable {
 ///
 ///        MyDisposable() {
 ///          var thing = new ThingThatRequiresCleanup();
-///          manageDisposer(() {
+///          getManagedDisposer(() {
 ///            thing.cleanUp();
 ///            return new Future(() {});
 ///          });
 ///        }
 ///      }
 ///
-/// Cleanup will then be automatically performed when the containing
+/// Cleanup will then be automatically performed when the parent
 /// object is disposed. If returning a future is inconvenient or
 /// otherwise undesirable, you may also return `null` explicitly.
 ///
@@ -198,14 +199,20 @@ class Disposable implements disposable_common.Disposable {
   void manageDisposable(disposable_common.Disposable disposable) =>
       _disposable.manageDisposable(disposable);
 
+  @deprecated
   @override
   void manageDisposer(disposable_common.Disposer disposer) =>
       _disposable.manageDisposer(disposer);
 
   @override
+  SimpleDisposable getManagedDisposer(disposable_common.Disposer disposer) =>
+      _disposable.getManagedDisposer(disposer);
+
+  @override
   void manageStreamController(StreamController controller) =>
       _disposable.manageStreamController(controller);
 
+  @deprecated
   @override
   void manageStreamSubscription(StreamSubscription subscription) =>
       _disposable.manageStreamSubscription(subscription);
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -19,7 +19,6 @@ import 'package:meta/meta.dart';
 import 'package:w_common/func.dart';
 
 import 'package:w_common/src/common/disposable.dart' as disposable_common;
-import 'package:w_common/src/common/disposable_manager.dart';
 
 class _InnerDisposable extends disposable_common.Disposable {
   Func<Future<Null>> onDisposeHandler;
@@ -205,7 +204,8 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageDisposer(disposer);
 
   @override
-  ManagedDisposer getManagedDisposer(disposable_common.Disposer disposer) =>
+  disposable_common.ManagedDisposer getManagedDisposer(
+          disposable_common.Disposer disposer) =>
       _disposable.getManagedDisposer(disposer);
 
   @override
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -205,7 +205,7 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageDisposer(disposer);
 
   @override
-  SimpleDisposable getManagedDisposer(disposable_common.Disposer disposer) =>
+  ManagedDisposer getManagedDisposer(disposable_common.Disposer disposer) =>
       _disposable.getManagedDisposer(disposer);
 
   @override
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -21,25 +21,45 @@ import 'package:meta/meta.dart';
 import 'package:w_common/src/common/disposable_manager.dart';
 import 'package:w_common/src/common/managed_stream_subscription.dart';
 
-// ignore: one_member_abstracts
-abstract class _Disposable {
-  Future<Null> dispose();
-}
-
-class InternalDisposable implements _Disposable {
+class InternalDisposable implements SimpleDisposable {
   Disposer _disposer;
+  final Completer<Null> _didDispose = new Completer<Null>();
+  bool _isDisposing = false;
 
   InternalDisposable(this._disposer);
 
   @override
   Future<Null> dispose() {
-    var disposeFuture = _disposer != null ? _disposer() : null;
-    _disposer = null;
-    if (disposeFuture == null) {
-      return new Future.value();
+    if (isDisposed) {
+      return null;
     }
-    return disposeFuture.then((_) => null);
+    if (_isDisposing) {
+      return didDispose;
+    }
+    _isDisposing = true;
+
+    var disposeFuture = _disposer != null
+        ? (_disposer() ?? new Future(() {}))
+        : new Future(() {});
+
+    return disposeFuture.then((_) {
+      _disposer = null;
+      _didDispose.complete();
+      _isDisposing = false;
+    });
   }
+
+  @override
+  Future<Null> get didDispose => _didDispose.future;
+
+  @override
+  bool get isDisposed => _didDispose.isCompleted;
+
+  @override
+  bool get isDisposedOrDisposing => isDisposed || isDisposing;
+
+  @override
+  bool get isDisposing => _isDisposing;
 }
 
 /// A [Timer] implementation that exposes a [Future] that resolves when a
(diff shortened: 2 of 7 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/src/browser/disposable_browser.dart
```diff
@@ -68,14 +69,14 @@ class _InnerDisposable extends disposable_common.Disposable {
...
-///          manageDisposer(() {
+///          getManagedDisposer(() {
 ///            thing.cleanUp();
 ///            return new Future(() {});
 ///          });
 ///        }
 ///      }
 ///
-/// Cleanup will then be automatically performed when the containing
+/// Cleanup will then be automatically performed when the parent
```
aaronstgeorge-wf: I went through and made sure that there was consistent use of `parent` vs `containing` object, not really sure which one I actually prefer. I chose `parent` simply because it was used more. From the perspective of the AST a managed object could definitely be thought of a child object, but `containing` maybe seems more natural for the disposable pattern. I'm on the fence.

lib/src/common/disposable_manager.dart
```diff
@@ -186,3 +257,30 @@ abstract class LeakFlagger {
 /// the object managing it is disposed, the future will complete with
 /// an instance of this exception.
 class ObjectDisposedException implements Exception {}
+
+/// Used to invoke, and remove references to, a [Disposer] before disposal of the
+/// parent object.
+abstract class SimpleDisposable {
```
aaronstgeorge-wf: There's probably a better name for this...

todbachman-wf: Now that you are exporting this class, let's come up with a better name!

lib/src/browser/disposable_browser.dart
```diff
@@ -198,14 +199,20 @@ class Disposable implements disposable_common.Disposable {
   void manageDisposable(disposable_common.Disposable disposable) =>
       _disposable.manageDisposable(disposable);
 
+  @deprecated
   @override
   void manageDisposer(disposable_common.Disposer disposer) =>
```
todbachman-wf: Add a doc comment identifying the migration path now that this method is deprecated.

aaronstgeorge-wf: The doc comment should be inherited from [here](https://github.com/Workiva/w_common/pull/47/files#diff-3758fc2d32c5ee729b97395b0dd3e885R41).

lib/src/common/disposable.dart
```diff
@@ -21,25 +21,45 @@ import 'package:meta/meta.dart';
...
-    _disposer = null;
-    if (disposeFuture == null) {
-      return new Future.value();
+    if (isDisposed) {
+      return null;
     }
-    return disposeFuture.then((_) => null);
+    if (_isDisposing) {
+      return didDispose;
+    }
```
bencampbell-wf: 🔮 Could this could be simplified? I wonder if `didDispose` future already resolved would a return of its value return on the current or next future?

```dart
if (isDisposedOrDisposing) {
    return didDispose;
}
```

lib/src/browser/disposable_browser.dart
```diff
@@ -198,14 +199,20 @@ class Disposable implements disposable_common.Disposable {
...
+  SimpleDisposable getManagedDisposer(disposable_common.Disposer disposer) =>
+      _disposable.getManagedDisposer(disposer);
+
+  @override
   void manageStreamController(StreamController controller) =>
       _disposable.manageStreamController(controller);
 
+  @deprecated
   @override
   void manageStreamSubscription(StreamSubscription subscription) =>
```
todbachman-wf: You've deprecated this, but I don't see what it has been deprecated in favor of.

lib/src/common/disposable.dart
```diff
@@ -21,25 +21,45 @@ import 'package:meta/meta.dart';
...
+    _isDisposing = true;
+
+    var disposeFuture = _disposer != null
+        ? (_disposer() ?? new Future(() {}))
+        : new Future(() {});
```
bencampbell-wf: 🔮 Why not set a default value on construction? Seems like it would make more sense then this dense expression as well as be safe for future changes.

lib/src/common/disposable.dart
```diff
@@ -21,25 +21,45 @@ import 'package:meta/meta.dart';
...
   InternalDisposable(this._disposer);
 
   @override
   Future<Null> dispose() {
-    var disposeFuture = _disposer != null ? _disposer() : null;
-    _disposer = null;
-    if (disposeFuture == null) {
-      return new Future.value();
+    if (isDisposed) {
+      return null;
```
bencampbell-wf: 🌵 What happens when a consumer awaits a null value? I think you should be returning a `Future`:

```diff
-      return null;
+      return Future.value(null);

File: lib/src/browser/disposable_browser.dart (hunk 5 above)
Comment (aaronstgeorge-wf) [inline lib/src/browser/disposable_browser.dart]: The doc comment with migration instructions should be inherited from [here](https://github.com/Workiva/w_common/blob/master/lib/src/common/disposable_manager.dart#L57)

File: lib/src/common/disposable_manager.dart (hunk 2 above)
Comment (bencampbell-wf) [inline lib/src/common/disposable_manager.dart]: `DisposablePrimitive`?

File: lib/src/browser/disposable_browser.dart (hunk 5 above)
Comment (aaronstgeorge-wf) [inline lib/src/browser/disposable_browser.dart]: It should have been deprecated in a previous PR, it was just overlooked.

File: lib/src/common/disposable_manager.dart (hunk 2 above)
Comment (todbachman-wf) [inline lib/src/common/disposable_manager.dart]: Also the doc comment is narrow than what the purpose of this class appears to be.

File: lib/src/common/disposable.dart (hunk 7 above)
Comment (aaronstgeorge-wf) [inline lib/src/common/disposable.dart]: What's wrong with awaiting null :trollface: ? 

This was done because that's what disposable does https://github.com/Workiva/w_common/blame/master/lib/src/common/disposable.dart#L280. @georgelesica-wf any particular reason? 

I originally wanted to just hand back a full `Disposable` under a limited interface, but that could screw with logging :( so I replicated the behavior for `InternalDisposable`.

File: lib/src/common/disposable_manager.dart (hunk 2 above)
Comment (aaronstgeorge-wf) [inline lib/src/common/disposable_manager.dart]: @todbachman-wf I'm not really following. The doc comment seems to be exactly what it's used for  from a consumer perspective. We use it internally as well but I did that just to get rid of the `_Disposable` class. Wondering how you thought this might be expanded?

File: lib/src/common/disposable.dart (hunk 4 above)
Comment (aaronstgeorge-wf) [inline lib/src/common/disposable.dart]: Maybe, @georgelesica-wf was there any reason this wasn't done originally in `Disposable`?

File: lib/src/common/disposable.dart (hunk 7 above)
Comment (bencampbell-wf) [inline lib/src/common/disposable.dart]: That is in an async block so the null return value will be returned as the value in a `Future` instance. What if this dispose method is used in a sync block:

```dart
internalDisposable.dispose().then((_) => print('I will throw because the dispose method returned null'));
```

aaronstgeorge-wf: Ahhh :) I didn't think about that being in a async block 👍

lib/src/common/disposable_manager.dart
todbachman-wf: It's the fact we're using it in a different context internally that bothers me. These doc comments are for us as well as for consumers.

todbachman-wf: I'd be happy with `DisposablePrimitive`

lib/src/browser/disposable_browser.dart
bencampbell-wf: The deprecation should belong on the interface alongside the documentation.

aaronstgeorge-wf: It's [deprecated there too](https://github.com/Workiva/w_common/pull/47/files#diff-3758fc2d32c5ee729b97395b0dd3e885R63). I found if it was not deprecated in both places I woulden't get analysis warnings in my IDE.

bencampbell-wf: Bizarre.

todbachman-wf: Do you see the doc comment inherited from the interface in your IDE @aaronstgeorge-wf ?

lib/src/common/disposable.dart
```diff
@@ -26,20 +26,41 @@ abstract class _Disposable {
   Future<Null> dispose();
 }
 
-class InternalDisposable implements _Disposable {
+class InternalDisposable implements _Disposable, ManagedDisposer {
```
todbachman-wf: Hmm. It seems like our InternalDisposable is just a ManagedDisposer. What if we just replace it with ManagedDisposer?

lib/src/browser/disposable_browser.dart
aaronstgeorge-wf: Yes I do @todbachman-wf

lib/src/common/disposable.dart
aaronstgeorge-wf: Are you saying rather than export an interface and implementing here, just create the implementation and export it?

todbachman-wf: Yep, that's exactly what I'm thinking.

aaronstgeorge-wf: Yep, I thought about that briefly. `ManagedDisposer` seemed a little bit like something a consumer shouldn't be able to construct, but that's not really a big deal. I'll go ahead and do that.

lib/src/common/disposable.dart
```diff
@@ -26,19 +26,50 @@ abstract class _Disposable {
   Future<Null> dispose();
 }
 
-class InternalDisposable implements _Disposable {
+/// Used to invoke, and remove references to, a [Disposer] before disposal of the
+/// parent object.
+class ManagedDisposer implements _Disposable {
```
todbachman-wf: What do you think, is there any value still to the `_Disposable` class?

lib/src/common/disposable.dart
```diff
@@ -26,19 +26,50 @@ abstract class _Disposable {
...
-    _disposer = null;
-    if (disposeFuture == null) {
-      return new Future.value();
+    if (isDisposedOrDisposing) {
+      return didDispose;
     }
-    return disposeFuture.then((_) => null);
+    _isDisposing = true;
+
+    var disposeFuture = _disposer() ?? new Future(() {});
```
todbachman-wf: I'm not sure you still need the null-aware operator here. Do you?

lib/src/common/disposable.dart
aaronstgeorge-wf: I think it sill provides value. It provides a common interface between `ManagedDisposer` and `Disposable`, this allows us to store both items in (and type) the `_internalDisposables` list in `Disposer`. We could make `Disposable` implement `ManagedDisposer`, I had that earlier today and it led to some confusion  exactly  the object was and how to document it.

lib/src/common/disposable.dart
aaronstgeorge-wf: It's possible that the call to `_disposer` returns `null` rather than a `Future<Null>`, we actually have some tests that trigger that.

lib/src/common/disposable.dart
todbachman-wf: I had searched through the code and thought we were only ever putting `ManagedDisposer`s in the `_internalDisposable` collection, but I must have missed the one case we're putting a `Disposable` in there. 🤷‍♂️

bencampbell-wf:
+10

Things Tested
-------------
- [x] CI Passed

bencampbell-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass
- [ ] Release Notes

Merging into master.
@Workiva/release-management-p

rmconsole-wf:
+1 from RM

PR #55, opened 2017-07-28
WP-4861 Add manageAndReturnDisposable method

srinivasdhanwada-wf:
### Description
Injecting dependencies into an object through named parameters has become a common, useful pattern as it allows the ability to specify these dependencies only if necessary and have a fallback that is constructed for you. 

In the case of disposables, the recommended pattern is to manage the disposable _only if_ the object creates a new instance. If a disposable gets passed into the object, the object should not manage the disposable as its lifetime might last longer. In order to accomplish this, the object must check if a dependency was passed in, and if not create one and manage it. There are several ways to accomplish this, but all involve conditional uses of `manageDisposable()` which start to add up as more dependencies are allowed: 
```dart
class MyDisposable extends Disposable {
  // This object also extends disposable
  MyObject _internal;

  MyDisposable({MyObject optional}) {
    // If optional is injected, we should not manage it.
    // If we create our own internal reference we should manage it.
    _internal = optional ?? new MyObject();
    if(optional == null) {
      manageDisposable(_internal);
    }
  }

  // ...
}
```

### Changes
Add a `getManagedDisposable` method that will manage the disposable passed in and return it back to the consumer to use further. 
```dart
class MyDisposable extends Disposable {
  // This object also extends disposable
  MyObject _internal;

  MyDisposable({MyObject optional}) {
    // If optional is injected, we should not manage it.
    // If we create our own internal reference we should manage it.
    _internal = optional ??
        getManagedDisposable(new MyObject());
  }

  // ...
}
```

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@Workiva/web-platform-pp 
@Workiva/rich-app-platform-pp

### DIFF
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -191,6 +191,11 @@ class Disposable implements disposable_common.Disposable {
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
   @override
+  disposable_common.Disposable manageAndReturnDisposable(
+          disposable_common.Disposable disposable) =>
+      _disposable.manageAndReturnDisposable(disposable);
+
+  @override
   Completer<T> manageCompleter<T>(Completer<T> completer) => _disposable
       .manageCompleter(completer);
 
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -212,7 +212,7 @@ typedef Future<dynamic> Disposer();
 /// without explicit reference to [Disposable]. To do this, we use
 /// composition to include the [Disposable] machinery without changing
 /// the public interface of our class or polluting its lifecycle.
-class Disposable implements _Disposable, DisposableManagerV5, LeakFlagger {
+class Disposable implements _Disposable, DisposableManagerV6, LeakFlagger {
   static bool _debugMode = false;
   static Logger _logger;
 
@@ -437,6 +437,15 @@ class Disposable implements _Disposable, DisposableManagerV5, LeakFlagger {
     });
 
     return managedStreamSubscription;
+  }
+
+  @mustCallSuper
+  @override
+  Disposable manageAndReturnDisposable(Disposable disposable) {
+    _throwOnInvalidCall('manageAndReturnDisposable', 'disposable', disposable);
+    manageDisposable(disposable);
+
+    return disposable;
   }
 
   @mustCallSuper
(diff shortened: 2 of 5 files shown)


aviary-wf:
## Raven
<details>

</details>

lib/src/browser/disposable_browser.dart
```diff
@@ -204,6 +204,11 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageDisposer(disposer);
 
   @override
+  disposable_common.Disposable getManagedDisposable(
```
georgelesica-wf: I would prefer to call this something like `manageAndReturnDisposable` or `manageDisposableV2`. We should also deprecate the existing `manageDisposable` since there is really no good reason to have both.

If breaking changes weren't so difficult to do I would say just update the existing method.

evanweible-wf: Doesn't `getManagedDisposable()` align with other similar methods that we have like `getManagedDisposer()` and `getManagedDelayedFuture()`?

georgelesica-wf: But in those cases the method is acting as a factory, in this case it's just passing its argument through.

srinivasdhanwada-wf: That makes sense to me, I can change it to `manageAndReturnDisposable`.

lib/src/browser/disposable_browser.dart
```diff
@@ -204,6 +204,11 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageDisposer(disposer);
 
   @override
+  disposable_common.Disposable manageAndReturnDisposable(
+          disposable_common.Disposable disposable) =>
+      _disposable.manageAndReturnDisposable(disposable);
```
evanweible-wf: alphabetize this now that it has been renamed

lib/src/common/disposable.dart
```diff
@@ -373,6 +373,15 @@ class Disposable implements _Disposable, DisposableManagerV5, LeakFlagger {
 
   @mustCallSuper
   @override
+  Disposable manageAndReturnDisposable(Disposable disposable) {
```
evanweible-wf: sort this alphabetically

lib/src/browser/disposable_browser.dart
srinivasdhanwada-wf: This is done

lib/src/common/disposable.dart
srinivasdhanwada-wf: This is also done

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
  - passing CI w/ tests exercising new functionality
- [ ] All unit tests pass

Merging into master.

PR #66, opened 2017-10-19
WP-5391 Add "disposal imminent" step prior to disposal

evanweible-wf:
### Description

There are 2 issues being addressed here:

1. The existing logic around awaitable futures (futures registered via `awaitBeforeDispose`) only waits for the set of futures available when `dispose()` is called – but ignores additional futures registered between calling `dispose()` and when all other awaitable futures have completed.
1. Currently, `isDisposing=true` as soon as `dispose()` is called. However, actual disposal doesn't begin until after the awaitable futures complete. This means that the `Disposable` API is unusable sooner than it needs to be because they check for `isDisposingOrDisposed`.

### Changes

- Use a `while` loop to continuously wait for registered futures until there are no more.
- Switch to an internal enum to manage the state of the Disposable instance
- Add a `Future<Null> willDispose();` lifecycle hook that can be implemented by extenders of `Disposable` – this hook gets called as soon as `dispose()` is called but before disposal actually begins. Potential use cases are to allow instances to cancel pending async tasks, do some cleanup, or kick start some actions that result in awaited futures completing so that disposal can actually occur.
- Add an "awaiting disposal" state that is entered as soon as `dispose()` is called. In this state, all of the `Disposable` APIs are still usable and it is perfectly valid to continue processing async tasks. The "disposing" state is not entered until `willDispose()` and all futures registered via `awaitBeforeDispose()` complete.
- Add an `bool isOrWillBeDisposed` getter that is `true` as soon as `dispose()` is called and remains true forever.
- Deprecate the other bool getters, as `isOrWillBeDisposed` should be sufficient for public consumers. All they really need to know is if disposal has been requested.

### Semantic Versioning

- [ ] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -19,13 +19,20 @@ import 'package:meta/meta.dart';
 import 'package:w_common/func.dart';
 
 import 'package:w_common/src/common/disposable.dart' as disposable_common;
+import 'package:w_common/src/common/disposable_state.dart';
 
 class _InnerDisposable extends disposable_common.Disposable {
   Func<Future<Null>> onDisposeHandler;
+  Func<Future<Null>> willDisposeHandler;
 
   @override
   Future<Null> onDispose() {
     return onDisposeHandler();
+  }
+
+  @override
+  Future<Null> willDispose() {
+    return willDisposeHandler();
   }
 }
 
@@ -41,15 +48,15 @@ class _InnerDisposable extends disposable_common.Disposable {
 /// default implementations and flexibility since it does not occupy
 /// a spot in the class hierarchy.
 ///
-/// Helper methods, such as [manageStreamSubscription] allow certain
-/// cleanup to be automated. Managed subscriptions will be automatically
-/// canceled when [dispose] is called on the object.
+/// Helper methods, such as [listenToStream] allow certain cleanup to be
+/// automated. Managed subscriptions will be automatically canceled when
+/// [dispose] is called on the object.
 ///
 ///      class MyDisposable extends Object with Disposable {
 ///        StreamController _controller = new StreamController();
 ///
 ///        MyDisposable(Stream someStream) {
-///          manageStreamSubscription(someStream.listen((_) => print('some stream')));
+///          listenToStream(someStream, (_) => print('some stream'));
 ///          manageStreamController(_controller);
 ///        }
 ///
@@ -141,9 +148,11 @@ class Disposable implements disposable_common.Disposable {
   @override
   bool get isDisposed => _disposable.isDisposed;
 
+  @deprecated
   @override
   bool get isDisposedOrDisposing => _disposable.isDisposedOrDisposing;
 
+  @deprecated
   @override
   bool get isDisposing => _disposable.isDisposing;
 
@@ -151,12 +160,22 @@ class Disposable implements disposable_common.Disposable {
   bool get isLeakFlagSet => _disposable.isLeakFlagSet;
 
   @override
-  Future<T> awaitBeforeDispose<T>(Future<T> future) => _disposable
-      .awaitBeforeDispose(future);
+  bool get isOrWillBeDisposed => _disposable.isOrWillBeDisposed;
+
+  @override
+  @protected
+  @visibleForTesting
+  DisposableState get state => _disposable.state;
+
+  @override
+  Future<T> awaitBeforeDispose<T>(Future<T> future) =>
+      _disposable.awaitBeforeDispose(future);
 
   @override
   Future<Null> dispose() {
-    _disposable.onDisposeHandler = this.onDispose;
+    _disposable
+      ..onDisposeHandler = this.onDispose
+      ..willDisposeHandler = this.willDispose;
     return _disposable.dispose().then((_) {
       // We want the description to be the runtime type of this
       // object, not the proxy disposable, so we need to reset
@@ -196,8 +215,8 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageAndReturnDisposable(disposable);
 
   @override
-  Completer<T> manageCompleter<T>(Completer<T> completer) => _disposable
-      .manageCompleter(completer);
+  Completer<T> manageCompleter<T>(Completer<T> completer) =>
+      _disposable.manageCompleter(completer);
 
   @override
   void manageDisposable(disposable_common.Disposable disposable) =>
@@ -223,9 +242,20 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageStreamSubscription(subscription);
 
   /// Callback to allow arbitrary cleanup on dispose.
-  @protected
   @override
+  @protected
   Future<Null> onDispose() async {
+    return null;
+  }
+
+  /// Callback to allow arbitrary cleanup as soon as disposal is requested (i.e.
+  /// [dispose] is called) but prior to disposal actually starting.
+  ///
+  /// Disposal will _not_ start before the [Future] returned from this method
+  /// completes.
+  @override
+  @protected
+  Future<Null> willDispose() async {
     return null;
   }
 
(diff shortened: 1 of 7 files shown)


aviary2-wf:
## Raven
<details>

</details>

lib/src/common/disposable.dart
```diff
@@ -256,9 +254,11 @@ class Disposable implements _Disposable, DisposableManagerV6, LeakFlagger {
...
   }
 
-  final Set<Future> _awaitableFutures = new HashSet<Future>();
-  Completer<Null> _didDispose = new Completer<Null>();
-  final Set<_Disposable> _internalDisposables = new HashSet<_Disposable>();
+  final _awaitableFutures = new HashSet<Future>();
+  final _didDispose = new Completer<Null>();
+  LeakFlag _leakFlag;
+  final _internalDisposables = new HashSet<_Disposable>();
+  bool _isDisposalImminent = false;
```
georgelesica-wf: This name is fine, but thinking of other suggestions just for discussion:

  * wasDisposeRequested
  * hasDisposeBegun
  * didBeginDispose
  * isDisposeInProgress

lib/src/common/disposable.dart
```diff
@@ -281,21 +281,54 @@ class Disposable implements _Disposable, DisposableManagerV6, LeakFlagger {
...
+  /// calls after disposal has been requested (via [dispose]) are rejected:
+  ///
+  ///     Response sendRequest() async {
+  ///       if (isDisposalImminentOrComplete) {
+  ///         throw new StateError(
+  ///             'sendRequest() cannot be called, object is disposing');
+  ///       }
+  ///       ...
+  ///     }
+  bool get isDisposalImminentOrComplete =>
```
georgelesica-wf: Out nomenclature is getting complicated. Technically this should be `isDisposalImminentOrInProgressOrComplete` or something crazy. I'm fine with considering naming tech debt to be hashed out later.

georgelesica-wf:
I think that this is a very reasonable way to address the problem we ran into. From what I can tell, anything that worked with the existing code will continue to work just fine, so that's nice.

lib/src/common/disposable.dart
jacehensley-wf: How bout `isOrWillBeDisposed`?

lib/src/common/disposable.dart
evanweible-wf: Yeah I was trying to find something that differentiated itself from `isDisposing`.. I wanted something that communicated the fact that disposal had been requested, but wasn't actually _in progress_ yet, since the instance is effectively unusable once disposal is in progress, but can be still be used up until that point.

evanweible-wf: I kind of like the "requested" nomenclature

lib/src/common/disposable.dart
evanweible-wf: Totally agree, I was struggling with that. Would it be worth scheduling a quick meeting to talk this through?

lib/src/common/disposable.dart
bencampbell-wf: If the poll is still open to suggestions I would submit `isAwaitingDisposal`.

lib/src/common/disposable.dart
```diff
@@ -331,13 +364,20 @@ class Disposable implements _Disposable, DisposableManagerV6, LeakFlagger {
...
+    _isDisposalImminent = true;
+
+    while (_awaitableFutures.isNotEmpty) {
+      final futures = _awaitableFutures.toList();
+      _awaitableFutures.clear();
+      await Future.wait(futures);
+    }
+
+    _isDisposing = true;
+    _isDisposalImminent = false;
```
bencampbell-wf: FWIW: Setting true/false flags for each state is easy to goof up. Is it better to have an enum mutually exclusive states and use that to set the current state of the disposable object?

evanweible-wf:
@Workiva/web-platform-pp @Workiva/rich-app-platform-pp I've pushed up a new commit with a slightly different approach and I'm looking for some feedback. In particular I'm wondering if people agree with my deprecation of `isDisposing` and `isDisposedOrDisposing` in favor of `isOrWillBeDisposed`, as well as the addition of a `willDispose()` lifecycle hook.

lib/src/browser/disposable_browser.dart
```diff
@@ -223,12 +242,23 @@ class Disposable implements disposable_common.Disposable {
...
   }
 
+  /// Callback to allow arbitrary cleanup as soon as disposal is requested (i.e.
+  /// [dispose] is called) but prior to disposal actually starting.
+  ///
+  /// Disposal will _not_ start before the [Future] returned from this method
+  /// completes.
+  @override
+  @protected
+  Future<Null> willDispose() async {
```
maxwellpeterson-wf: Should this be `onWillDispose` to follow the semantic pattern of the other lifecycle methods?

What is the intended use case of `willDispose` vs. `onDispose`?

evanweible-wf: I wondered that, too, I just thought it sounded weird to have "on" and "will" back to back.. but I suppose `onWillDispose` would better align with our `Experience` class.

There's an unfortunate amount of ambiguity around naming conventions for booleans, event streams, and overridable handlers (in our own stuff and when you consider things like React).

Most of the Experience streams start with `did`, but not every event is the result of something happening – e.g. what about an event for this scenario where we're about to do something?

We also have a `bool get didEnterIdleState` and an `Future<Null> get onEnterIdleState` on the Experience.

React just uses the overridable methods, but does it in the form of `componentWill` and `componentDid`.

`state_machine` states have `onEnter` and `onLeave` streams.

`Disposable.didDispose` is a `Future`..

¯\_(ツ)_/¯

---

I'll post a couple of example use cases for `willDispose` in a little bit – I'm using it in the w_module PR

evanweible-wf: I forgot that `LifecycleModule` has a bunch of `did` and `will` streams, so you're right that it definitely makes more sense for this to be `onWillDispose`

evanweible-wf:
@maxwellpeterson-wf here's an example use case: https://github.com/Workiva/w_module/pull/101/files#diff-c586827f6717f60db651d466b5ab2b5fR676

In `LifecycleModule`, `onWillDispose` is implemented so that the module can try to unload itself (a process which will emit events on streams that need to be intact) _before_ disposal starts. Before, this was not possible, because you could only listen to `didDispose` (too late), override `dispose()` (frowned upon, more complicated than it needs to be), or override `onDispose()` (again, too late, because disposal has already started).

maxwellpeterson-wf:
@jayudey-wf ready for you

lib/src/common/disposable.dart
```diff
@@ -331,13 +381,21 @@ class Disposable implements _Disposable, DisposableManagerV6, LeakFlagger {
     if (isDisposed) {
       return null;
     }
-    if (_isDisposing) {
+    if (isOrWillBeDisposed) {
```
jayudey-wf: should this be tested?

jayudey-wf:
testing in consumers

evanweible-wf:
Consumer testing with this branch of w_common and [this branch of w_module](https://github.com/Workiva/w_module/pull/101):
- [x] wdesk_sdk had a few failing tests, fixed here: https://github.com/Workiva/wdesk_sdk/pull/1883
- [x] home ✅ 
- [x] graph_app ✅ 
- [x] datatables ✅ 
- [x] docs ✅

maxwellpeterson-wf:
+10
- Tested in consumers
- CI passes

maxwellpeterson-wf:
QA +1
- [x] Dev +1's
  - Others have reviewed and approved the significant changes, I've refreshed my approval after the latest commit.
- [x] Tests created/updated 
- [x] All tests pass

rmconsole-wf:
+1 from RM

PR #81, opened 2018-03-05
WP-6173 Fixes for Dart2

robbecker-wf:
## Dart 2 (dev channel) Problems
 - There's a missing implementation of the `tick` method for `_ObservableTimer`. 
```
w_common git:(master) ✗ dda

::: dartanalyzer --fatal-warnings --strong lib/disposable.dart lib/invalidation_mixin.dart lib/cache.dart lib/disposable_browser.dart lib/func.dart lib/json_serializable.dart lib/w_common.dart test/unit/typedefs.dart test/unit/stubs.dart test/unit/disposable_common.dart tool/dev.dart
    Analyzing lib/disposable.dart, lib/invalidation_mixin.dart, lib/cache.dart, lib/disposable_browser.dart, lib/func.dart, lib/json_serializable.dart, lib/w_common.dart, test/unit/typedefs.dart, test/unit/stubs.dart, test/unit/disposable_common.dart, tool/dev.dart...
      error • Missing concrete implementation of getter 'Timer.tick' at lib/src/common/disposable.dart:83:7 • non_abstract_class_inherits_abstract_member_one
    1 error found.

Analysis failed.
```

# Changes
 - Implement tick, but for now just throw unsupported. This lets the library work under Dart 1 & 2.
 - Turn on Travis CI testing for both versions of Dart to make sure it works in both places.

### DIFF
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -115,10 +115,10 @@ class _ObservableTimer implements Timer {
 
   @override
   int get tick {
-      // TODO: Once fully transitioned to Dart 2 just return
-      // the tick value from our internal timer
-      // return _timer.tick;
-      throw new UnsupportedError('Timer.tick is unsupported');
+    // TODO: Once fully transitioned to Dart 2 just return
+    // the tick value from our internal timer
+    // return _timer.tick;
+    throw new UnsupportedError('Timer.tick is unsupported');
   }
 }
 
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -112,6 +112,12 @@ class _ObservableTimer implements Timer {
 
   @override
   bool get isActive => _timer.isActive;
+
+  @override
+  int get tick => _timer.tick; // Dart2 only
+  //int get tick => 0; // Works for Dart1 & 2, but will this break something about timers?
+  // can we re-implement the tick value correctly? Should we inject our own zone handler for
+  // timer creation to call the complete method?
 }
 
 /// A class used as a marker for potential memory leaks.
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -114,10 +114,12 @@ class _ObservableTimer implements Timer {
   bool get isActive => _timer.isActive;
 
   @override
-  int get tick => _timer.tick; // Dart2 only
-  //int get tick => 0; // Works for Dart1 & 2, but will this break something about timers?
-  // can we re-implement the tick value correctly? Should we inject our own zone handler for
-  // timer creation to call the complete method?
+  int get tick {
+      // TODO: Once fully transitioned to Dart 2 just return
+      // the tick value from our internal timer
+      // return _timer.tick;
+      throw new UnsupportedError('Timer.tick is unsupported');
+  }
 }
 
 /// A class used as a marker for potential memory leaks.
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Files and directories created by pub
 .packages
-.pub/
+.pub
+.dart_tool
 build/
 packages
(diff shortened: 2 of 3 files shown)


aviary2-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

lib/src/common/disposable.dart
```diff
@@ -112,6 +112,12 @@ class _ObservableTimer implements Timer {
 
   @override
   bool get isActive => _timer.isActive;
+
+  @override
+  int get tick => _timer.tick; // Dart2 only
+  //int get tick => 0; // Works for Dart1 & 2, but will this break something about timers?
+  // can we re-implement the tick value correctly? Should we inject our own zone handler for
+  // timer creation to call the complete method?
```
robbecker-wf: @georgelesica-wf Thoughts? Under dart 1 a tick of always zero would be ok since it is not used. Under Dart 2, I'm not sure what always returning zero would break in Dart 2, if anything.

robbecker-wf: From the DAB Core library changes doc:

> The original Timer implementation ensured that all ticks of a periodic timer would execute. If the timer fell behind, it would try to catch up, but would never lose ticks. This behavior leads to big problems when a computer goes to sleep (or a Dart program isn't executed for a long time). When the Dart program is resumed, it will try to run *all* missed ticks. In Dart 2.x, it will skip the missed ticks. In return, the Timer class has a tick member that informs the program how many ticks should have passed:

```dart
/**
   * The number of durations preceding the most recent timer event.
   *
   * The value starts at zero and is incremented each time a timer event
   * occurs, so each callback will see a larger value than the previous one.
   *
   * If a periodic timer with a non-zero duration is delayed too much,
   * so more than one tick should have happened,
   * all but the last tick in the past are considered "missed",
   * and no callback is invoked for them.
   * The [tick] count reflects the number of durations that have passed and
   * not the number of callback invocations that have happened.
   */
  int get tick;
```

robbecker-wf: we should probably either implement tick correctly, or delegate it to the underlying timer.tick. If you can think of a way to do this that is compatible with Dart 1 and Dart 2 . I'm all ears.

lib/src/common/disposable.dart
```diff
@@ -112,6 +112,14 @@ class _ObservableTimer implements Timer {
 
   @override
   bool get isActive => _timer.isActive;
+
+  @override
+  int get tick {
```
evanweible-wf: So is `tick` even a field on `Timer` in Dart 1? (i.e. is this `@override` valid in Dart 1?)

robbecker-wf: No, the override results in a warning in Dart 1, but it still works.

robbecker-wf:
review @Workiva/team-missoula ?

lib/src/common/disposable.dart
```diff
@@ -112,6 +112,14 @@ class _ObservableTimer implements Timer {
...
+    // TODO: Once fully transitioned to Dart 2 just return
+    // the tick value from our internal timer
+    // return _timer.tick;
+    throw new UnsupportedError('Timer.tick is unsupported');
```
seanburke-wf: Won't this cause this not to work in dart 2 anyhow? Would it be possibly to try to access `_timer.tick` and throw if it throws?

robbecker-wf: Tried that. It is not possible to try to use it if it is there. I asked the Dart team and they suggested just throwing UnsupportedError.

patkujawa-wf:
QA +1 CI passed.

Quality Review Approval: +1

- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass
- [x] Unit test created/updated
- [x] Functional tests (skynet) pass or failures are justified


@Workiva/release-management-p for merge into master

PR #90, opened 2018-04-10
MSODP-263 add applyToItem method

aaronstgeorge-wf:
### Description

Working with consumption in `internal application X` recently has shown that it would be a helpful for the cache to have a get that works like a map get rather than a cache get. In other words the caching strategy would not be informed of the get and it would not count towards against the removal of the item being referenced from the cache. This can be very useful when one wants to mutate an item that is known to be in the cache.

### Changes

Added a method to the cache that will run a callback on an item in the cache without doing a get. As a bonus it will ensure that the events will not be put on the `didRemove` stream until the callback has finished running. To guard against the situation where a consumer chooses to dispose on removal of an object from the cache and applies a long running callback with quite a few awaits. In this circumstance it's possible that the object will be disposed before there callback finishes; which could lead to (in all likelihood benign) exceptions thrown.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
+
 import 'package:w_common/func.dart';
 import 'package:w_common/src/common/disposable.dart';
 
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -381,8 +381,9 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   ///
   /// This does not perform a [get] or [getAsync], and as a result, will not
   /// affect retention or removal of a [TIdentifier][TValue] pair from the cache.
-  /// If any callbacks are in fight on removal of [TIdentifier] [TValue] pair
-  /// [didRemove] will not event until callback has completed.
+  ///
+  /// Any [TIdentifier] [TValue] pair removals will wait for the Future returned by
+  /// the call to [callback] before emitting [didRemove] events.
   ///
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   Future<bool> applyToItem(
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -172,7 +172,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
   Future<TValue> get(TIdentifier id, Func<TValue> valueFactory) {
-    _log.finest("get id: $id");
+    _log.finest('get id: $id');
     _throwWhenDisposed('get');
     _cachingStrategy.onWillGet(id);
     // Await any pending cached futures
@@ -230,7 +230,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
   Future<TValue> getAsync(TIdentifier id, Func<Future<TValue>> valueFactory) {
-    _log.finest("getAsync id: $id");
+    _log.finest('getAsync id: $id');
     _throwWhenDisposed('getAsync');
     _cachingStrategy.onWillGet(id);
     // Await any pending cached futures
@@ -257,7 +257,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// the current [CachingStrategy].
   @mustCallSuper
   Future<Null> release(TIdentifier id) {
-    _log.finest("release id: $id");
+    _log.finest('release id: $id');
     _throwWhenDisposed('release');
     // Await any pending cached futures
     if (_cache.containsKey(id)) {
@@ -284,7 +284,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
   Future<Null> remove(TIdentifier id) {
-    _log.finest("remove id: $id");
+    _log.finest('remove id: $id');
     _throwWhenDisposed('remove');
     if (_cache.containsKey(id)) {
       _cachingStrategy.onWillRemove(id);
@@ -357,7 +357,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   ///     }
   Future<TValue> weakGet(TIdentifier id) {
     _throwWhenDisposed('weakGet');
-    _log.finest("weakGet id: $id");
+    _log.finest('weakGet id: $id');
     return _cache[id];
   }
 
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -86,7 +86,7 @@ class CachingStrategy<TIdentifier, TValue> {
 /// An abstraction over [Map] that helps avoid paying construction costs for
 /// expensive objects.
 ///
-/// Objects are created or existing ones returned with [get] or its async
+/// Objects are created or existing ones returned with [get] or its `async`
 /// equivalent [getAsync] and marked as eligible for removal with [release]. If
 /// the [TValue] stored requires some sort of destruction this should be done in
 /// a callback registered with the [didRemove] stream.
@@ -97,9 +97,11 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   final Logger _log = new Logger('w_common.Cache');
 
   /// The backing store for values in the [Cache].
-  Map<TIdentifier, Future<TValue>> _cache = <TIdentifier, Future<TValue>>{};
+  final Map<TIdentifier, Future<TValue>> _cache =
+      <TIdentifier, Future<TValue>>{};
 
-  Map<TIdentifier, bool> _isReleased = <TIdentifier, bool>{};
+  /// Whether a given identifier has been released.
+  final Map<TIdentifier, bool> _isReleased = <TIdentifier, bool>{};
 
   /// The current caching strategy, set at construction.
   final CachingStrategy<TIdentifier, TValue> _cachingStrategy;
@@ -143,9 +145,8 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   Iterable<TIdentifier> get keys =>
       _cache.keys.where((TIdentifier key) => !_isReleased[key]);
 
-  Future<Iterable<TValue>> get values => Future.wait(_cache.keys
-      .where((TIdentifier key) => !_isReleased[key])
-      .map((TIdentifier key) => _cache[key]));
+  Future<Iterable<TValue>> get values =>
+      Future.wait(keys.map((TIdentifier key) => _cache[key]));
 
   /// Does the [Cache] contain the given [TIdentifier]?
   ///
@@ -327,12 +328,11 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     return new Future.value();
   }
 
-  /// Run [callback] on [TValue] associated with [id].
+  /// Returns `true` and calls [callback] if there is a cached, unreleased
+  /// [TValue] associated with [id]; returns `false` otherwise.
   ///
-  /// If [TIdentifier] [TValue] pair found runs [callback] and returns true,
-  /// otherwise returns false. Will not run on released items and does not
-  /// perform a [get] or [getAsync]. As [get] or [getAsync] are not performed
-  /// calling [applyToItem] will not affect retention or removal of a
+  /// This does not perform a [get] or [getAsync], and as a result,
+  /// will not affect retention or removal of a
   /// [TIdentifier][TValue] pair from the cache.
   ///
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -152,19 +152,19 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// To be removed: 2.0.0
   ///
   /// This entry point is deprecated in favor of the more precisely named
-  /// [releasedKeys] and [nonReleasedKeys].
+  /// [releasedKeys] and [liveKeys].
   @deprecated
   Iterable<TIdentifier> get keys => _cache.keys;
 
   /// Keys that have not been released.
-  Iterable<TIdentifier> get nonReleasedKeys =>
+  Iterable<TIdentifier> get liveKeys =>
       _cache.keys.where((TIdentifier key) => !_isReleased[key]);
 
   /// Values that have not been released.
   ///
   /// To access a released value a [get] or [getAsync] should be used.
-  Future<Iterable<TValue>> get nonReleasedValues =>
-      Future.wait(nonReleasedKeys.map((TIdentifier key) => _cache[key]));
+  Future<Iterable<TValue>> get liveValues =>
+      Future.wait(liveKeys.map((TIdentifier key) => _cache[key]));
 
   /// Keys that have been released but are not yet removed.
   Iterable<TIdentifier> get releasedKeys =>
@@ -178,9 +178,9 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// To be removed: 2.0.0
   ///
   /// This entry point is deprecated in favor of the more precisely named
-  /// [nonReleasedValues].
+  /// [liveValues].
   @deprecated
-  Future<Iterable<TValue>> get values => nonReleasedValues;
+  Future<Iterable<TValue>> get values => liveValues;
 
   /// Does the [Cache] contain the given [TIdentifier]?
   ///
@@ -189,7 +189,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// Deprecated: 1.12.0
   /// To be removed: 2.0.0
   ///
-  /// This entry point is deprecated in favor of using [nonReleasedKeys].contains
+  /// This entry point is deprecated in favor of using [liveKeys].contains
   /// or [releasedKeys].contains directly.
   @deprecated
   bool containsKey(TIdentifier id) {
@@ -315,10 +315,10 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// Release indicates that consuming code has finished with the [TIdentifier]
   /// [TValue] pair associated with [id]. This pair may be removed immediately or
   /// in time depending on the [CachingStrategy]. Access to the pair will be
-  /// blocked immediately, i.e. the [nonReleasedValues] getter won't report this
+  /// blocked immediately, i.e. the [liveValues] getter won't report this
   /// item, even if the pair isn't immediately removed from the cache. A [get] or
-  /// [getAsync] must be performed to mark the item as not eligible for removal
-  /// and live in the cache again.
+  /// [getAsync] must be performed to mark the item as ineligible for removal and
+  /// live in the cache again.
   ///
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
--- a/lib/src/common/cache/cache.dart
+++ b/lib/src/common/cache/cache.dart
@@ -82,7 +82,13 @@ class CachingStrategy<TIdentifier, TValue> {
   void onWillRemove(TIdentifier id) {}
 }
 
-/// Maintains a reference to a given [TValue] associated with a [TIdentifier].
+/// An abstraction over [Map] that helps avoid paying construction costs for
+/// expensive objects.
+///
+/// Objects are created or existing ones returned with [get] or its async
+/// equivalent [getAsync] and marked as eligible for removal with [release]. If
+/// the [TValue] stored requires some sort of destruction this should be done in
+/// a callback registered with the [didRemove] stream.
 ///
 /// References are retained for the lifecycle of the instance of the [Cache],
 /// unless explicitly removed.
@@ -91,6 +97,8 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
 
   /// The backing store for values in the [Cache].
   Map<TIdentifier, Future<TValue>> _cache = <TIdentifier, Future<TValue>>{};
+
+  Map<TIdentifier, bool> _isReleased = <TIdentifier, bool>{};
 
   /// The current caching strategy, set at construction.
   final CachingStrategy<TIdentifier, TValue> _cachingStrategy;
@@ -131,16 +139,19 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
       _didUpdateController.stream;
 
-  Iterable<TIdentifier> get keys => _cache.keys;
+  Iterable<TIdentifier> get keys =>
+      _cache.keys.where((TIdentifier key) => !_isReleased[key]);
 
-  Future<Iterable<TValue>> get values => Future.wait(_cache.values);
+  Future<Iterable<TValue>> get values => Future.wait(_cache.keys
+      .where((TIdentifier key) => !_isReleased[key])
+      .map((TIdentifier key) => _cache[key]));
 
   /// Does the [Cache] contain the given [TIdentifier]?
   ///
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   bool containsKey(TIdentifier id) {
     _throwWhenDisposed('determine if identifier is cached');
-    return _cache.containsKey(id);
+    return _cache.containsKey(id) && !_isReleased[id];
   }
 
   /// Returns a value from the cache for a given [TIdentifier].
@@ -175,6 +186,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     _log.finest('get id: $id');
     _throwWhenDisposed('get');
     _cachingStrategy.onWillGet(id);
+    _isReleased[id] = false;
     // Await any pending cached futures
     if (_cache.containsKey(id)) {
       return _cache[id].then((TValue value) async {
@@ -232,6 +244,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   Future<TValue> getAsync(TIdentifier id, Func<Future<TValue>> valueFactory) {
     _log.finest('getAsync id: $id');
     _throwWhenDisposed('getAsync');
+    _isReleased[id] = false;
     _cachingStrategy.onWillGet(id);
     // Await any pending cached futures
     if (_cache.containsKey(id)) {
@@ -255,10 +268,21 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   ///
   /// The decision of whether or not to actually remove the value will be up to
   /// the current [CachingStrategy].
+  ///
+  /// Release indicates that consuming code has finished with the [TIdentifier]
+  /// [TValue] pair associated with [id]. This pair may be removed immediately or
+  /// in time depending on the [CachingStrategy]. Access to the pair will be
+  /// blocked immediately, i.e. the [keys] and [values] getters won't report this
+  /// item, even if the pair isn't immediately removed from the cache. A [get] or
+  /// [getAsync] must be performed to mark the item as not eligible for removal
+  /// and live in the cache again.
+  ///
+  /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
   Future<Null> release(TIdentifier id) {
     _log.finest('release id: $id');
     _throwWhenDisposed('release');
+    _isReleased[id] = true;
     // Await any pending cached futures
     if (_cache.containsKey(id)) {
       _cachingStrategy.onWillRelease(id);
@@ -286,6 +310,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   Future<Null> remove(TIdentifier id) {
     _log.finest('remove id: $id');
     _throwWhenDisposed('remove');
+    _isReleased.remove(id);
     if (_cache.containsKey(id)) {
       _cachingStrategy.onWillRemove(id);
       final removedValue = _cache.remove(id);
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,65 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  ///
+  /// [weakGet] can be thought of sort of like a weak reference when working with
+  /// a garbage collector. [weakGet] won't prevent removal of a [TValue]
+  /// [TIdentifier] pair from the [Cache], [get] or [getAsync] will. In other
+  /// words, it's a get that behaves like `Map`'s does rather than the way
+  /// [Cache]'s does. This can be convenient, for example, when mutating an
+  /// object that is known to be in the cache.
+  ///
+  /// This ideally shouldn't be used to create a reference with a lifetime that
+  /// will exceed the time an item exists in the cache. For example this would be
```
georgelesica-wf: s/an item/the item/

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,65 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  ///     await cache.release('thing');                                                ┛┃
+  ///     // Await a random future to ensure that the callback registered on didRemove  ┃
+  ///     // is called.                                                                 ┃
+  ///     await new Future((){});                                                       ┃
+  ///                                                                                   ┃
+  ///     // This will throw because the instance associated with 'thing' has been      ┃
+  ///     // removed from the cache and disposed on the didRemove callback.             ┃
+  ///     danglingReference.manageDisposable(new Disposable());                         ┋
+  ///
+  /// An usual consumption pattern would look something like:
```
smaifullerton-wk: `A usual`

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,65 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  /// [TIdentifier] pair from the [Cache], [get] or [getAsync] will. In other
+  /// words, it's a get that behaves like `Map`'s does rather than the way
+  /// [Cache]'s does. This can be convenient, for example, when mutating an
+  /// object that is known to be in the cache.
+  ///
+  /// This ideally shouldn't be used to create a reference with a lifetime that
+  /// will exceed the time an item exists in the cache. For example this would be
+  /// a bad consumption pattern:
+  ///
+  /// (when a reference created with [weakGet] would be alive and when an item
```
smaifullerton-wk: Maybe something like "The delineation to the right of the code indicates when a reference created with [weakGet] would be alive and when an item will be in the cache" would be more clear

lib/src/common/cache/cache.dart
```diff
@@ -169,6 +172,7 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
   @mustCallSuper
   Future<TValue> get(TIdentifier id, Func<TValue> valueFactory) {
+    _log.finest("get id: $id");
```
evanweible-wf: #nit single qutoes

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,66 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     return new Future.value();
   }
 
+  /// Provides access to a value stored in the cache without informing the
+  /// [CachingStrategy] of a get.
+  ///
+  /// [weakGet] can be thought of sort of like a weak reference when working with
```
greglittlefield-wf: I think the term "weak" here is misleading (I'm going to be PRing a `WeakMap` implementation soon that uses true weak references).

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,66 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  ///                                                                                  ┃
+  ///     var danglingReference = await cache.weakGet('thing');                        ┃┓
+  ///                                                                                  ┃┃
+  ///     await cache.release('thing');                                                ┛┃
+  ///     // Await a random future to ensure that the callback registered on didRemove  ┃
+  ///     // is called.                                                                 ┃
+  ///     await new Future((){});                                                       ┃
+  ///                                                                                   ┃
+  ///     // This will throw because the instance associated with 'thing' has been      ┃
+  ///     // removed from the cache and disposed on the didRemove callback.             ┃
```
greglittlefield-wf: Isn't this a problem with the consumption pattern of disposing on removal, and not with the caching class itself?

It seems like a better approach would be to not dispose the object automatically if there are cases where you might access it after it gets removed. Or, alternatively, keep that automatic disposal in place and check whether the object is disposed before interacting with it.

greglittlefield-wf: Is this the use-case where you'd like to have the behavior, or just an example? I'd like to get more insight into the problem you're trying to solve so I can understand the need for this behavior better.

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,66 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     return new Future.value();
   }
 
+  /// Provides access to a value stored in the cache without informing the
+  /// [CachingStrategy] of a get.
+  ///
+  /// [weakGet] can be thought of sort of like a weak reference when working with
+  /// a garbage collector. [weakGet] won't prevent removal of a [TValue]
+  /// [TIdentifier] pair from the [Cache], [get] or [getAsync] will. In other
```
greglittlefield-wf: > [weakGet] won't prevent removal of a [TValue] [TIdentifier] pair from the [Cache], [get] or [getAsync] will

Could you elaborate on how these prevent removal?

lib/src/common/cache/cache.dart
aaronstgeorge-wf: This is intended to be an example of a bad consumption pattern that should be avoided. 

### For a little bit of clarification on the problem trying to be solved:

In TDC we have a least-recently-used loaded experience cache. When an experience goes on screen it's loaded and put into the cache. When it goes off screen it's marked as `released` meaning that's it's eligible for being unloaded (which will happen when it is removed from the cache). The idea is to prevent the user having to wait for load every time an experience is briefly scrolled off screen then back on while still ensuring that the experience will be removed from memory at some point. 

There are several cases where something needs to call a method on one of these loaded experiences.  This was done by doing a cache `get` for the experience then calling the method on the returned experience. Calling one of these methods implies that you expect the experience to be loaded and onscreen, if you're calling one of these methods on something that is potentially unloaded and off screen it's an error. Some errant go code was calling one of these methods at an improper time (after the same code has informed TDC that it's ok for the experience to be unloaded i.e `release`, but not long enough for the experience to be unloaded). The cache `get` required to get at the experience removed it from the pool of experiences marked as eligible for removal, meaning that the experience would never unloaded.

This API is an attempt to create a way  get an item from the cache, that is expected to be there, to mutate it or access some property that won't lead to retaining it in the cache. The errant code calling a method at the wrong time also needs to be fixed, but we believe it would be a more robust consumption pattern to have calls that come at the wrong time to get a `null` rather than leading to unintentional retention.

### As for the consumption pattern: 
Usually you wouldn't hide an object behind a cache unless it's pretty heavy, so I would think that disposing on removal would be a pretty usual consumption pattern.

lib/src/common/cache/cache.dart
aaronstgeorge-wf: If there's going to be a `WeakMap` I agree :/ this name will be confusing. `getIfCached` isn't bad maybe `getExisting`?

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +301,66 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  ///       }
+  ///
+  ///       void removeAThing(String id) {
+  ///         _cache.release(id);
+  ///       }
+  ///     }
+  Future<TValue> weakGet(TIdentifier id) {
+    _throwWhenDisposed('weakGet');
+    _log.finest('weakGet id: $id');
+    return _cache[id];
```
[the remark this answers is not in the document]
aaronstgeorge-wf: Yes this is intended to work like a `Map` get rather than a `Cache` get so if the value isn't there you'll get a `null` value. I can write a test for that.

lib/src/common/cache/cache.dart
aaronstgeorge-wf: Can do.

lib/src/common/cache/cache.dart
aaronstgeorge-wf: screen shot 2018-04-12 at 1 59 24 pm

Right now `weakGet` (or whatever we end up calling it) could be used to get access to released but not yet removed items without going through a `get`. `release` should mean that from a consumption perspective it's fine that this item is removed from memory. Having a side channel that provides access to things that are `released` but not yet removed seems like a leaky abstraction.

georgelesica-wf: Instead of a getter you could provide a method that will run a function against an item in the cache without re-upping its "cached" status.

```dart
void applyToItem(TIdentifier id, void callback(TValue value));
```

aaronstgeorge-wf: I like that. Semantically it's a little cleaner and I think it naturally warns people away from the bad consumption pattern detailed in the doc comment.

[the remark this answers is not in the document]
aaronstgeorge-wf: Everything is stored in the cache as a `Future<TValue>` so I'm not sure what we can do to get around the future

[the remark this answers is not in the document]
georgelesica-wf: Oh, I see, then it should probably be `return _cache[id] ?? Future.null()`.

georgelesica-wf: I guess the callback would actually be `void callback(Future<TValue> value)`, but yeah...

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +327,25 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     return new Future.value();
   }
 
+  /// Run [callback] on [TValue] associated with [id].
```
greglittlefield-wf: #nit this Is a bit vague for the first sentence of a Dart doc comment, and the second paragraph could be [condensed](https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-brevity) a bit.
```dart
/// Returns `true` and calls [callback] if there is a cached, unreleased
/// [TValue] associated with [id]; returns `false` otherwise.
///
/// This does not perform a [get] or [getAsync], and as a result,
/// will not affect retention or removal of a
/// [TIdentifier][TValue] pair from the cache.
///
/// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
```

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +348,99 @@ void main() {
         expect(() => cache.release(cachedId), throwsStateError);
       });
     });
+
+    group('keys', () {
+      test('should not provide access to released keys', () {
+        // Ensure that cached item is not removed
```
greglittlefield-wf: This comment would be well-suited for the `reason` argument to `expectAsync1`.

lib/src/common/cache/cache.dart
```diff
@@ -81,14 +83,24 @@ class CachingStrategy<TIdentifier, TValue> {
...
 ///
 /// References are retained for the lifecycle of the instance of the [Cache],
 /// unless explicitly removed.
 class Cache<TIdentifier, TValue> extends Object with Disposable {
+  final Logger _log = new Logger('w_common.Cache');
+
   /// The backing store for values in the [Cache].
   Map<TIdentifier, Future<TValue>> _cache = <TIdentifier, Future<TValue>>{};
 
+  Map<TIdentifier, bool> _isReleased = <TIdentifier, bool>{};
```
greglittlefield-wf: #nit Could use a doc comment

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +348,99 @@ void main() {
...
+    group('values', () {
+      test('should not provide access to released values', () async {
+        // Ensure that cached item is not removed
+        cache.didRemove
+            .listen(expectAsync1((CacheContext context) {}, count: 0));
+
+        expect(await cache.values, contains(cachedValue));
+        // ignore: unawaited_futures
+        cache.release(cachedId);
+        expect((await cache.values).contains(cachedValue), isFalse);
```
greglittlefield-wf: #nit Could use `isNot` for a better error message:
```dart
expect(await cache.values, isNot(contains(cachedValue)));
```

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +348,99 @@ void main() {
...
+        expect(calbackRan, isFalse);
+      });
+
+      test('should run callback if item is in the cache', () {
+        var calbackRan = false;
+        cache.applyToItem(cachedId, (_) {
+          calbackRan = true;
+        });
+
+        expect(calbackRan, isTrue);
```
greglittlefield-wf: Can we also verify the argument passed to the callback?

lib/src/common/cache/cache.dart
```diff
@@ -250,9 +269,21 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   ///
   /// The decision of whether or not to actually remove the value will be up to
   /// the current [CachingStrategy].
+  ///
+  /// Release indicates that consuming code has finished with the [TIdentifier]
+  /// [TValue] pair associated with [id]. This pair may be removed immediately or
+  /// in time depending on the [CachingStrategy]. Access to the pair will be
+  /// blocked immediately, i.e. the [keys] and [values] getters won't report this
```
greglittlefield-wf: > Access to the pair will be blocked immediately

Is this not a behavioral breakage?

lib/src/common/cache/cache.dart
```diff
@@ -128,16 +140,19 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   Stream<CacheContext<TIdentifier, TValue>> get didUpdate =>
       _didUpdateController.stream;
 
-  Iterable<TIdentifier> get keys => _cache.keys;
+  Iterable<TIdentifier> get keys =>
+      _cache.keys.where((TIdentifier key) => !_isReleased[key]);
 
-  Future<Iterable<TValue>> get values => Future.wait(_cache.values);
+  Future<Iterable<TValue>> get values => Future.wait(_cache.keys
+      .where((TIdentifier key) => !_isReleased[key])
```
evanweible-wf: This logic is duplicated from the `keys` getter above, you could just reuse that here:

```dart
=> Future.wait(keys.map((TIdentifier key) => _cache[key]));
```

Unless you’d rather be explicit and not rely on the implementation of `keys` which could change.. I’m okay with it either way :)

lib/src/common/cache/cache.dart
evanweible-wf: They would know better as they’re the ones using it, but it seems to me that the previous description of this API:

> The decision of whether or not to actually remove the value will be up to the current [CachingStrategy].

..implies that it _could_ be removed immediately, in which case consumers would also be unable to access it without another `.get()` or `.getAsync()` call. If that was the case, then I don’t see a behavioral difference because consumers would have had to guard against this scenario anyway.

lib/src/common/cache/cache.dart
georgelesica-wf: I like this because, in my mind, the implementation of `keys` and the implementation of `values` should always be consistent with one another.

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +327,25 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  /// otherwise returns false. Will not run on released items and does not
+  /// perform a [get] or [getAsync]. As [get] or [getAsync] are not performed
+  /// calling [applyToItem] will not affect retention or removal of a
+  /// [TIdentifier][TValue] pair from the cache.
+  ///
+  /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
+  bool applyToItem(TIdentifier id, void callback(Future<TValue> value)) {
+    _log.finest('applyToItem id: $id');
+    _throwWhenDisposed('applyToItem');
+    if (_isReleased[id] != false) {
```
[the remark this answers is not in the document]
georgelesica-wf: Oh wait, never mind, I see what this is doing, clever!

lib/src/common/cache/cache.dart
greglittlefield-wf: Sounds good to me 👍. Just making sure it was considered

aaronstgeorge-wf:
I've noticed a few consumption patterns issues in `internal-app-x` I'd like to take a little more time to investigate those before moving forward with this.  ATM I am consumed in support work so it will be a little while before I can circle back around.

lib/src/common/cache/cache.dart
```diff
@@ -278,11 +351,16 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
     _throwWhenDisposed('remove');
+    _isReleased.remove(id);
     if (_cache.containsKey(id)) {
       _cachingStrategy.onWillRemove(id);
       final removedValue = _cache.remove(id);
       return removedValue.then((TValue value) async {
+        if (_applyToItemCallBacks[id] != null) {
+          await Future.wait(_applyToItemCallBacks[id]);
+        }
         await _cachingStrategy.onDidRemove(id, value);
```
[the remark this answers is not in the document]
aaronstgeorge-wf: [line 356](https://github.com/Workiva/w_common/pull/90/files#diff-1fdd0bfc9ee195804c63d431b8409625R356) removes it from the `_isReleased` map synchronously, and then [line 388](https://github.com/Workiva/w_common/pull/90/files#diff-1fdd0bfc9ee195804c63d431b8409625R388) will short circuit the apply.

lib/src/common/cache/cache.dart
```diff
@@ -250,9 +311,21 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
   ///
   /// The decision of whether or not to actually remove the value will be up to
   /// the current [CachingStrategy].
+  ///
+  /// Release indicates that consuming code has finished with the [TIdentifier]
+  /// [TValue] pair associated with [id]. This pair may be removed immediately or
+  /// in time depending on the [CachingStrategy]. Access to the pair will be
+  /// blocked immediately, i.e. the [nonReleasedValues] getter won't report this
+  /// item, even if the pair isn't immediately removed from the cache. A [get] or
+  /// [getAsync] must be performed to mark the item as not eligible for removal
```
georgelesica-wf: "ineligible"

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +372,37 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+    }
+    final callbackFuture = callback(_cache[id]);
+    if (callbackFuture is Future<Null>) {
+      _applyToItemCallBacks.putIfAbsent(id, () => <Future<Null>>[]);
+      _applyToItemCallBacks[id].add(callbackFuture);
+      callbackFuture.then((_) {
+        _applyToItemCallBacks[id].remove(id);
+        if (_applyToItemCallBacks[id].isEmpty) {
+          _applyToItemCallBacks.remove(id);
+        }
```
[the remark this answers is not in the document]
georgelesica-wf: Is there a way you can test this?

georgelesica-wf:
+1 looks fine to me, we should have a couple more people review just to guard against anything subtle

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
aaronstgeorge-wf: I think [this test](https://github.com/Workiva/w_common/pull/90/files#diff-71e7ef0509871bfd43a03a17b3e52771R452) should cover it. The more I think about it the less worried about a concurrent modification issue, there can only be one thing actually executing at one time.

lib/src/common/cache/cache.dart
```diff
@@ -327,23 +372,35 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
     _log.finest('applyToItem id: $id');
     _throwWhenDisposed('applyToItem');
     if (_isReleased[id] != false) {
-      return false;
+      return new Future.value(false);
+    }
+    final callbackFuture = callback(_cache[id]);
+    if (callbackFuture is Future<Null>) {
+      _applyToItemCallBacks.putIfAbsent(id, () => <Future<Null>>[]);
+      _applyToItemCallBacks[id].add(callbackFuture);
```
greglittlefield-wf: From above: if `<Null>` is dropped or it's made dynamic, it should also be dropped here and on the typing for `_applyToItemCallBacks`.
```dart
if (callbackFuture is Future) {
  _applyToItemCallBacks.putIfAbsent(id, () => <Future>[]);
  _applyToItemCallBacks[id].add(callbackFuture);

File: lib/src/common/cache/cache.dart [hunk 21]
```diff
@@ -327,23 +372,35 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+      callbackFuture.then((_) {
```
greglittlefield-wf: What happens when the future completes with an error? This should be probably be a `.whenComplete`, right? 

There should at least be tests around this.

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
greglittlefield-wf: The problem isn't two things executing at the same time, it's mutating a list while it's being iterated.

This case looks pretty safe from that case, though. If you want more peace of mind, a good way I've found to prevent concurrent modification errors is to make a copy of the iterable first, like so:
```dart
await Future.wait(_applyToItemCallBacks[id].toList());
```

---

Other gotchas around `Future.wait` to look out for that I found while thinking about this case:

Using `Future.wait` with a `.map` in which modifications can happen during iteration can be risky, especially in Dart 2 once async functions run their first part synchronously:
```dart
main() {
  var values = [1, 2];
  Future.wait(values.map((value) {
    return new Future.sync(() {
      values.remove(value); // Concurrent modification during iteration: Instance of 'MappedListIterable'.
    });
  }));
}
```
```dart
main() {
  var values = [1, 2];
  Future.wait(values.map((value) {
    values.remove(value); // Concurrent modification during iteration: Instance of 'MappedListIterable'.
    return new Future(() {});
  }));
}
```
```dart
// Dart 2
main() {
  var values = [1, 2];
  Future.wait(values.map((value) async {
    values.remove(value); // Concurrent modification during iteration: Instance of 'MappedListIterable'.
    await new Future(() {});
  }));
}
```

lib/src/common/cache/cache.dart
```diff
@@ -327,23 +372,35 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
   ///
-  /// If [TIdentifier] [TValue] pair found runs [callback] and returns true,
-  /// otherwise returns false. Will not run on released items and does not
-  /// perform a [get] or [getAsync]. As [get] or [getAsync] are not performed
-  /// calling [applyToItem] will not affect retention or removal of a
-  /// [TIdentifier][TValue] pair from the cache.
+  /// This does not perform a [get] or [getAsync], and as a result, will not
+  /// affect retention or removal of a [TIdentifier][TValue] pair from the cache.
+  /// If any callbacks are in fight on removal of [TIdentifier] [TValue] pair
+  /// [didRemove] will not event until callback has completed.
```
greglittlefield-wf: This could probably be made clearer and its own paragraph.

```dart
///
/// Any [TIdentifier] [TValue] pair removals will wait for the Future returned by
/// the call to [callback] before emitting [didRemove] events.
```

Also, perhaps I missed this somewhere, but why is this the desired behavior?

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
greglittlefield-wf: Sounds like this would be good to have a test for.

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +372,37 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
     return new Future.value();
   }
 
+  /// Returns `true` and calls [callback] if there is a cached, unreleased
+  /// [TValue] associated with [id]; returns `false` otherwise.
+  ///
+  /// This does not perform a [get] or [getAsync], and as a result, will not
+  /// affect retention or removal of a [TIdentifier][TValue] pair from the cache.
+  /// If any callbacks are in fight on removal of [TIdentifier] [TValue] pair
+  /// [didRemove] will not event until callback has completed.
```
greglittlefield-wf: This could probably be made clearer and its own paragraph.

```dart
///
/// Any [TIdentifier] [TValue] pair removals will wait for the Future returned by
/// the call to [callback] before emitting [didRemove] events.
```

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +372,37 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  ///
+  /// If the [Cache] [isOrWillBeDisposed] then a [StateError] is thrown.
+  Future<bool> applyToItem(
+      TIdentifier id, Future<Null> callback(Future<TValue> value)) {
```
greglittlefield-wf: I think `Future<Null>` is unnecessarily restrictive here, and that it should just be 
```dart
Future callback(Future<TValue> value)
```

Actually, can we just make it `dynamic`, so that consumers can pass in synchronous functions?

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +372,37 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+    _throwWhenDisposed('applyToItem');
+    if (_isReleased[id] != false) {
+      return new Future.value(false);
+    }
+    final callbackFuture = callback(_cache[id]);
+    if (callbackFuture is Future<Null>) {
+      _applyToItemCallBacks.putIfAbsent(id, () => <Future<Null>>[]);
+      _applyToItemCallBacks[id].add(callbackFuture);
+      callbackFuture.then((_) {
+        _applyToItemCallBacks[id].remove(id);
```
greglittlefield-wf: Shouldn't this be `.remove(callbackFuture)`?

Might be a good idea to expose a `@visibleForTesting ` version of `_applyToItemCallBacks` to ensure it's getting managed as expected.

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
greglittlefield-wf: Also, what if a `get` comes in while you're waiting for this to finish? Would it be possible to get an event sequence that looks like:
```
<item not in cache>
item added to cache
didUpdate
apply with callback that takes awhile
item removed from cache
apply callback still hasn't finished
item added to the cache
didUpdate
apply callback finishes
didRemove
<item is still in cache>
```
Could this unexpected sequence of events cause issues for consumers that may not be expecting it?

lib/src/common/cache/cache.dart
aaronstgeorge-wf: Yes it absolutely should.

aaronstgeorge-wf:
>Some comments, mostly around the _applyToItemsCallbacks additions. Could you provide some context on why those changes are needed?
>
>I'm not saying I necessarily disagree with those additions, but I'd like to understand why this increased complexity is necessary.

I would consider this a nice to have rather a necessity. If a consumer chooses to dispose on removal of an object from the cache and applies a long running callback with quite a few `awaits` it's possible that the object will be disposed before there callback finishes; which could lead to (in all likelihood benign) exceptions thrown.

I would be fine with removing `_applyToItemsCallbacks` additions if you feel the "juice is not worth the squeeze" to quote @maxwellpeterson-wf.

aaronstgeorge-wf:
>Also, what if a `get` comes in while you're waiting for this to finish? Would it be possible to get an 
event sequence that looks like:
```
<item not in cache>
item added to cache
didUpdate
apply with callback that takes awhile
item removed from cache
apply callback still hasn't finished
item added to the cache
didUpdate
apply callback finishes
didRemove
<item is still in cache>
```
>Could this unexpected sequence of events cause issues for consumers that may not be expecting it?

Great point! This would not be a issue for the consumption patterns I have seen.

lib/src/common/cache/cache.dart
aaronstgeorge-wf: https://github.com/Workiva/w_common/pull/90#issuecomment-387116341

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
aaronstgeorge-wf: Yeah what I meant to say was that the only way we would get mutation of a list while it's being iterated with the current structure is if we had multiple things executing at one time.

lib/src/common/cache/cache.dart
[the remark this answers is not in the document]
aaronstgeorge-wf: > Sounds like this would be good to have a test for.

[Here's one!](https://github.com/Workiva/w_common/pull/90/files#diff-71e7ef0509871bfd43a03a17b3e52771R450)

lib/src/common/cache/cache.dart
```diff
@@ -128,16 +150,55 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+  /// To be removed: 2.0.0
+  ///
+  /// This entry point is deprecated in favor of using [liveKeys].contains
+  /// or [releasedKeys].contains directly.
+  @deprecated
   bool containsKey(TIdentifier id) {
-    _throwWhenDisposed('determine if identifier is cached');
-    return _cache.containsKey(id);
+    _throwWhenDisposed('containsKey');
+    return keys.contains(id) || releasedKeys.contains(id);
```
evanweible-wf: this should be `liveKeys` instead of `keys`, right?

lib/src/common/cache/cache.dart
```diff
@@ -294,6 +376,47 @@ class Cache<TIdentifier, TValue> extends Object with Disposable {
...
+    _throwWhenDisposed('applyToItem');
+    if (_isReleased[id] != false) {
+      return new Future<bool>.value(false);
+    }
+
+    final callBackResult = callback(_cache[id]);
+    if (callBackResult is Future) {
+      // In this case we're only interested in the computation being done or not
+      // done, not in the result
+      final errorlessCallbackResult = callBackResult.catchError((_) {});
```
evanweible-wf: Is there ever a scenario where you would want to log an error thrown by the callback?

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +343,222 @@ void main() {
...
+        cache.didRemove.listen(expectAsync1((CacheContext context) {},
+            count: 0, reason: 'Ensure that cached item is not removed'));
+
+        expect(cache.releasedKeys, isNot(contains(cachedId)));
+        await cache.release(cachedId);
+        expect(cache.releasedKeys, contains(cachedId));
+      });
+    });
+
+    group('nonReleasedKeys', () {
```
evanweible-wf: liveKeys

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +343,222 @@ void main() {
...
+        cache.didRemove.listen(expectAsync1((CacheContext context) {},
+            count: 0, reason: 'Ensure that cached item is not removed'));
+
+        expect(cache.liveKeys, contains(cachedId));
+        await cache.release(cachedId);
+        expect(cache.liveKeys.contains(cachedId), isFalse);
+      });
+    });
+
+    group('nonReleasedValues', () {
```
evanweible-wf: liveValues

test/unit/browser/cache/cache_test.dart
```diff
@@ -342,6 +343,222 @@ void main() {
...
+
+          expect(callbackRan, isFalse);
+        });
+
+        test('should return true', () async {
+          expect(await cache.applyToItem(cachedId, (_) {}), isTrue);
+        });
+      });
+
+      group('should not event didRemove stream until callback has completed',
```
evanweible-wf: add event to? emit event from?

lib/src/common/cache/cache.dart
greglittlefield-wf: If so, it might be a good idea to use `.whenComplete` to allow this to propagate as an uncaught async error, which can be picked up by a `runZoned`'s `onError` handler and logged.

greglittlefield-wf:
> I would consider this a nice to have rather a necessity. If a consumer chooses to dispose on removal of an object from the cache and applies a long running callback with quite a few awaits it's possible that the object will be disposed before there callback finishes; which could lead to (in all likelihood benign) exceptions thrown.

@aaronstgeorge-wf That makes sense, thanks for the context there! 

Why would the exceptions likely be benign in that case, or in the event sequence I mentioned? Couldn't unknowingly having disposed objects in the cache potentially cause serious issues within and application? 

Perhaps there should be logic to not emit `onDidRemove` when a get comes in while waiting for the `_applyToItemCallBacks ` to finish?

lib/src/common/cache/cache.dart
greglittlefield-wf: Yup, I totally agree that the cache shouldn't care if the callback errors. The problem with the current code is that calling `.catchError` prevents that error from propagating to a zone handler.

I think that we should use `whenComplete` instead, which has the same behavior as the cache currently has, while allowing the error to propagate to consumer zones instead of being swallowed by `catchError`.
```dart
final completer = new Completer();
final errorlessCallbackResult = completer.future;
callBackResult.whenComplete(completer.complete);
```

---

Illustration of this behavior (which I wrote to make sure I understood it correctly 😄):

```dart
import 'dart:async';

main() {
  runZoned(() {
    handleFutureWhenComplete(new Future.error('error 1'));
    handleFutureCatchError(new Future.error('error 2'));
  }, onError: (error) {
    print('Zone onError: $error');
  });  
}

void handleFutureWhenComplete(Future future) {
  var completer = new Completer();
  var errorlessCallbackFuture = completer.future;
  
  future.whenComplete(completer.complete);
  
  errorlessCallbackFuture.then((_) {
    print('errorlessCallbackFuture (whenComplete) completed');
  }); 
}

void handleFutureCatchError(Future future) {
  var errorlessCallbackFuture = future.catchError((_) {});  
  errorlessCallbackFuture.then((_) {
    print('errorlessCallbackFuture (catchError) completed');
  }); 
}
```
Prints:
```
Zone onError: error 1
errorlessCallbackFuture (catchError) completed
errorlessCallbackFuture (whenComplete) completed
```

aaronstgeorge-wf: Ah yeah, but we're returning the original future to the callers zone on line [415](https://github.com/Workiva/w_common/pull/90/files/971bf535b30057780368fe9d8322bd6825665334#diff-1fdd0bfc9ee195804c63d431b8409625R415) not the new future we created with `catchError`.

So what we have is more like this: 

```dart
import 'dart:async';

main() {
  runZoned(() {
    handleFutureWhenComplete(new Future.error('error 1'));
    handleFutureCatchError(new Future.error('error 2'));
  }, onError: (error) {
    print('Zone onError: $error');
  });  
}

void handleFutureWhenComplete(Future future) {
  var completer = new Completer();
  var errorlessCallbackFuture = completer.future;
  
  future.whenComplete(completer.complete);
  
  errorlessCallbackFuture.then((_) {
    print('errorlessCallbackFuture (whenComplete) completed');
  }); 
}

void handleFutureCatchError(Future future) {
  var errorlessCallbackFuture = future.catchError((_) {});
  errorlessCallbackFuture.then((_){
     print('errorlessCallbackFuture (catchError) completed');
  });
  future.then((_) {
    print('future completed');
  }); 
}
```
Prints
```
Zone onError: error 1
errorlessCallbackFuture (catchError) completed
Zone onError: error 2
errorlessCallbackFuture (whenComplete) completed
```

aaronstgeorge-wf: That being said, maybe the `handleFutureWhenComplete` version is more idiomatic ¯\\\_(ツ)_/¯

aaronstgeorge-wf:
> _Edit: Ohhh, or is it not an issue since it'll be a different `TValue` instance that is returned?_

Yep :) @greglittlefield-wf

smaifullerton-wk:
### QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] All unit tests pass

@Workiva/release-management-p 🌻

PR #98, opened 2018-08-01
AF-1910 fixes for dart 1 and dart 2 compatibility

robbecker-wf:
### Description
Fixes to allow dart 1 and dart 2 compatibility at the same time.

### Changes
- CI changes (travis and Dockerfile) to use dartanalyzer and dartfmt directly
- test on chrome instead of dartium
- 1 small Type fix
- some changes for mockito under dart 2

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes


aviary-wf:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/information-security/aviary#configuring-aviary) coverage of security relevant code
---
_Questions or Comments? Reach out on HipChat: InfoSec Forum._

test/unit/browser/generated_browser_tests.dart
```diff
@@ -4,9 +4,12 @@ library test.unit.browser.generated_browser_tests;
 // Generated by `pub run dart_dev gen-test-runner -d test/unit/browser -e Environment.browser --no-genHtml`
```
robbecker-wf: only format changes

test/unit/browser/invalidation_mixin_test.dart
```diff
@@ -46,7 +46,7 @@ void main() {
       });
 
       test('calls validate, eventually', () {
-        Future onValidation = thing.invalidate();
+        Future<ValidationStatus> onValidation = thing.invalidate();
```
robbecker-wf: explicit Type here for Dart 2 inference later on at `.then(expectAsync1((ValidationStatus status) {`

.travis.yml
```diff
@@ -3,16 +3,14 @@ sudo: required
 
 dart:
   - 1.24.3
-  - stable
```
georgelesica-wf: Should we leave stable for once 2.0 is released?

.travis.yml
```diff
@@ -3,16 +3,14 @@ sudo: required
...
-  - test: --platform vm
-  - test: --platform dartium
-  - install_dartium: true
+  - dartfmt: true
+  - dartanalyzer: lib example tool test/unit
 
 script:
   - pub publish --dry-run # Validate package
   - pub run dart_dev gen-test-runner --check
-  - pub run dart_dev format --check
```
georgelesica-wf: At least right now, this takes configuration from dev.dart, so should we keep it?

pubspec.yaml
```diff
@@ -1,26 +1,27 @@
...
 
 dev_dependencies:
-  coverage: ^0.7.2
-  dart_dev: ^1.8.0
+  build_runner: ">=0.6.0 <1.0.0"
+  build_test: ">=0.9.0 <1.0.0"
+  build_web_compilers: ">=0.2.0 <1.0.0"
+  dart_dev: ^1.9.5
   dart_style: ^1.0.8
-  dartdoc: ^0.13.0
```
georgelesica-wf: Noice

.travis.yml
robbecker-wf: I'd like it to run on dart 2 today ... so maybe we should just specify 2.0.0-dev.69.5 and then update it to 2.0.0 / stable once 2.0 lands?

georgelesica-wf:
+1

.travis.yml
robbecker-wf: either way. I was experimenting with trying to get rid of the dart_style dependency .. but ended up restoring it anyway... so no reason why we can't restore this to dart_dev format.

lib/src/common/disposable.dart
```diff
@@ -417,9 +417,8 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
 
     if (_debugMode) {
       stopwatch.stop();
-      _logger
-          .info('$runtimeType $hashCode took ${stopwatch.elapsedMicroseconds /
-          1000000.0} seconds to dispose');
+      var t = stopwatch.elapsedMicroseconds / 1000000.0;
+      _logger.info('$runtimeType $hashCode took $t seconds to dispose');
```
robbecker-wf: broke this into 2 lines so formatting wouldn't be different between dart 1 and dart 2 🤷‍♂️

.travis.yml
```diff
@@ -11,6 +11,6 @@ dart_task:
 script:
   - pub publish --dry-run # Validate package
   - pub run dart_dev gen-test-runner --check
-  - tool/test.sh
+  - xvfb-run --server-args="-screen 0 1024x768x24" tool/test.sh
```
robbecker-wf: :doh: of course!

.travis.yml
evanweible-wf: Yeah I was just thinking that once Dart 2 stable lands, we'd come back and update these to:
```yaml
dart:
  - 1.24.3
  - stable
  - dev
```

evanweible-wf: Or maybe we leave the `dev` channel out at that point.

.travis.yml
```diff
@@ -1,18 +1,18 @@
...
+  - test: --platform chrome
 
 script:
-  - pub publish --dry-run # Validate package
-  - pub run dart_dev gen-test-runner --check
-  - pub run dart_dev format --check
-  - pub run dart_dev analyze
+  #- xvfb-run --server-args="-screen 0 1024x768x24" tool/test.sh
   #- pub run dart_dev coverage --no-html
   - bash <(curl -s https://codecov.io/bash) -f coverage/coverage.lcov
```
evanweible-wf: @robbecker-wf did you ever figure out if using `dart_task` results in the `script` section being skipped?

dart_test.yaml
```diff
@@ -0,0 +1,27 @@
...
+- chrome
+- vm
+
+paths:
+- test/unit/browser
+- test/unit/vm
+
+presets:
+  dart2:
+    exclude_tags: "no-dart2"
```
evanweible-wf: There aren't any tests that are tagged with `no-dart2` so you could remove this preset if you want.

Dockerfile
```diff
@@ -14,8 +14,8 @@ ARG GIT_MERGE_BRANCH
 WORKDIR /build/
 ADD . /build/
 RUN echo "Starting the script sections" && \
-		pub get --packages-dir && \
-		xvfb-run -s '-screen 0 1024x768x24' pub run dart_dev test --pub-serve --web-compiler=dartdevc -p chrome -p vm && \
-		echo "Script sections completed"
+	pub get --packages-dir && \
+	xvfb-run -s '-screen 0 1024x768x24' tool/test.sh && \
```
evanweible-wf: Why run tests in workiva build as well as travis CI?

georgelesica-wf: I'm not sure, maybe RM asked us to?

.travis.yml
georgelesica-wf: It does not.

robbecker-wf: I just discovered that, surprisingly, having a script section and a dart_task section results in only running the script section. So, in this file, you're not actually running tests, format, or analyzer. 😳

.travis.yml
georgelesica-wf: I think we should do all three

.travis.yml
georgelesica-wf: :tableflip:

Dockerfile
aaronlademann-wf: AFAIK, all RM cares about is that a `pub get` is run so that the dependency check can take place.

Dockerfile
```diff
@@ -14,8 +14,8 @@ ARG GIT_MERGE_BRANCH
 WORKDIR /build/
 ADD . /build/
 RUN echo "Starting the script sections" && \
-		pub get --packages-dir && \
-		xvfb-run -s '-screen 0 1024x768x24' pub run dart_dev test --pub-serve --web-compiler=dartdevc -p chrome -p vm && \
-		echo "Script sections completed"
+	pub get --packages-dir && \
```
robbecker-wf: I believe we should be able to remove `--packages-dir`

robbecker-wf:
@georgelesica-wf ready for review

pubspec.yaml
```diff
@@ -1,26 +1,22 @@
 name: w_common
-description: General utilities for Dart projects.
 version: 1.13.0
+description: General utilities for Dart projects.
 author: Workiva Client Platform Team <team-clientplatform@workiva.com>
 homepage: https://www.github.com/Workiva/w_common
 
 environment:
-  sdk: '>=1.12.0 <2.0.0'
+  sdk: '>=1.12.0 <3.0.0'
```
georgelesica-wf: Is this lower bound actually correct? I don't think it is because I don't think we got parametric methods until after 1.12.0. Maybe we should just set this to 1.24.3 since Dart 1 is basically deprecated at this point anyway, so there's no good reason to use an old version of it.

test/unit/browser/cache/cache_test.dart
```diff
@@ -204,31 +225,37 @@ void main() {
         var childCache = new Cache(stubCachingStrategy);
         await childCache.get(cachedId, () => cachedValue);
         await childCache.remove(cachedId);
-        verify(stubCachingStrategy.onDidRemove(cachedId, cachedValue));
+
+        expect(stubCachingStrategy.onDidRemoveCalled == 1, isTrue);
```
georgelesica-wf: Better error messages if you do `expect(stubCachingStrategy.onDidRemoveCalled, 1);`, same goes for the others.

test/unit/browser/cache/cache_test.dart
```diff
@@ -563,14 +594,72 @@ void main() {
   });
 }
 
-class MockCachingStrategy extends Mock
-    implements CachingStrategy<String, Object> {
-  MockCachingStrategy() {
-    when(this.onDidGet(typed(any), typed(any)))
-        .thenReturn(new Future.value(null));
-    when(this.onDidRelease(typed(any), typed(any), typed(any)))
-        .thenReturn(new Future.value(null));
```
georgelesica-wf: If you use `thenAnswer(() => new Future.value(null));` instead of `thenReturn...` this should work.

test/unit/browser/cache/caching_strategy_common_test.dart
```diff
@@ -1,3 +1,18 @@
+// Copyright 2018 Workiva Inc.
```
georgelesica-wf: Technically, I think these are supposed to be the years the file was modified in, so if it was created in 2016 and no modified until now it would be `2016, 2018` or if it was also modified in 2017 it would be `2016-2018`.

test/unit/browser/invalidation_mixin_test.dart
```diff
@@ -59,8 +60,10 @@ void main() {
 }
 
 class InvalidThing extends InvalidationMixin {
-  StreamController _onValidate = new StreamController.broadcast();
-  Stream get onValidate => _onValidate.stream;
+  final StreamController<ValidationStatus> _onValidate =
```
georgelesica-wf: #nit don't need the type annotation here

pubspec.yaml
robbecker-wf: Makes sense. I'm ok updating it if you are.

georgelesica-wf:
+1

.travis.yml
```diff
@@ -1,17 +1,14 @@
...
+addons:
+  chrome: stable
 
 dart:
   - 1.24.3
-dart_task:
-  - test: --platform vm
-  - test: --platform dartium
-  - install_dartium: true
+  - 2.0.0
```
todbachman-wf: This could just be `stable` now.

todbachman-wf:
+10. Code looks good and CI passed.

PR #104, opened 2018-12-10
implement idle task queue using requestIdleCallback

toddbeckman-wf:
### Description

Adds library-level [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) support in Dart.

Also adds an `IdleTaskQueue`, which can be used to queue general tasks that should only be run during idle time so that consumers don't need to write their own redundant yielding logic.

### Semantic Versioning

- [ ] **Patch**
- [x] **Minor**
* [ ] Major

### Testing/QA

TODO: Standard testing might not work here because in CI `requestIdleCallback` never fires because there is no idle frame. We need another test strategy for this one.

#### Testing Suggestion

My apologies to OSSers who can't see these things:

- PR to put AI reporting in idle task queue: https://github.com/Workiva/app_intelligence_dart/pull/763
- PR to put polyfill in Wdesk: https://github.com/Workiva/wdesk_sdk/pull/3124
- Test branch https://github.com/Workiva/text_doc_client/pull/7145

Use the test branch to verify that reporting (using networking) is not done during idle frames. Caveat: in debug mode, the console reporter is utilized, and that does *not* utilize idle tasks for obvious reasons.

### Code Review

@Workiva/app-frameworks

### DIFF
--- a/lib/idle_task_queue.dart
+++ b/lib/idle_task_queue.dart
@@ -1,0 +1,78 @@
+@JS()
+library idle_task_queue;
+
+import 'dart:async';
+import 'dart:html';
+import 'dart:js';
+
+import 'package:js/js.dart';
+import 'package:meta/meta.dart';
+
+/// This calls the [requestIdleCallback] API directly.
+///
+/// Note that `requestIdleCallbackPolyfill.js` or another polyfill like it must
+/// be included in order to use this API in browsers such as IE 11 or Edge which
+/// do not support it.
+///
+/// Most tasks could be added using [enqueueIdleTask] instead.
+@JS()
+external num requestIdleCallback(Function callback, [Map options]);
+
+/// A handler which will be called during idle time.
+typedef void IdleTaskHandler<T>(T input);
+
+@immutable
+class _IdleTask<T> {
+  final IdleTaskHandler<T> handler;
+  final T input;
+
+  _IdleTask(this.handler, this.input);
+}
+
+/// Adds the task to a queue which will be called during idle time.
+void enqueueIdleTask<T>(IdleTaskHandler<T> handler, T input) {
+  // For unsupported browsers when the polyfill is not available,
+  // we still want the task to be called eventually.
+  if (!_doesBrowserSupportIdleCallback) {
+    new Future.delayed(const Duration()).then((_) => handler(input));
+    return;
+  }
+
+  _taskQueue.add(new _IdleTask<T>(handler, input));
+
+  // Start running the queue if it has not been started.
+  // If there is a task, the queue is already running and doesn't
+  // need to be started again.
+  if (_currentTask == null) {
+    _currentTask = requestIdleCallback(
+      allowInterop(_runQueue),
+      <String, dynamic>{'timeout': 1000},
+    );
+  }
+}
+
+void _runQueue(IdleDeadline deadline) {
+  while (!deadline.didTimeout &&
+      deadline.timeRemaining() > 0 &&
+      _taskQueue.isNotEmpty) {
+    final task = _taskQueue.removeAt(0);
+    task.handler(task.input);
+  }
+
+  if (_taskQueue.isNotEmpty) {
+    _currentTask = requestIdleCallback(
+      allowInterop(_runQueue),
+      <String, dynamic>{
+        'timeout': 1000,
+      },
+    );
+  } else {
+    _currentTask = null;
+  }
+}
+
+bool get _doesBrowserSupportIdleCallback =>
+    context['requestIdleCallback'] != null;
+
+final List<_IdleTask> _taskQueue = [];
+num _currentTask;
--- a/lib/idle_task_queue.dart
+++ b/lib/idle_task_queue.dart
@@ -7,18 +7,15 @@ import 'dart:js';
 
 import 'package:meta/meta.dart';
 
-const Duration _immediately = const Duration();
-
 /// A handler which will be called during idle time.
-typedef void IdleTaskHandler<T>(T input);
+typedef FutureOr<R> IdleTaskHandler<R>();
 
 @immutable
-class _IdleTask<T> {
-  final IdleTaskHandler<T> handler;
-  final T input;
-  final Completer<Null> completer = new Completer<Null>();
+class _IdleTask<R> {
+  final IdleTaskHandler<R> handler;
+  final Completer<R> completer = new Completer<R>();
 
-  _IdleTask(this.handler, this.input);
+  _IdleTask(this.handler);
 }
 
 /// Adds the task to a queue which will be called during idle time.
@@ -26,16 +23,16 @@ class _IdleTask<T> {
 /// Returns the `Future` which will complete when the task has run.
 ///
 /// May complete with an error if the task throws when called.
-Future<Null> enqueueIdleTask<T>(IdleTaskHandler<T> handler, T input) {
+Future<R> enqueueIdleTask<R>(IdleTaskHandler<R> handler) {
   // For unsupported browsers when the polyfill is not available,
   // we still want the task to be called eventually.
   if (!_doesBrowserSupportIdleCallback) {
-    return new Future.delayed(_immediately).then((_) async {
-      handler(input);
+    return new Future.delayed(Duration.ZERO).then((_) async {
+      return handler();
     });
   }
 
-  final task = new _IdleTask<T>(handler, input);
+  final task = new _IdleTask<R>(handler);
   _taskQueue.add(task);
 
   // Start running the queue if it has not been started.
@@ -43,7 +40,7 @@ Future<Null> enqueueIdleTask<T>(IdleTaskHandler<T> handler, T input) {
   // need to be started again.
   if (_currentTask == null) {
     _currentTask = window.requestIdleCallback(
-      allowInterop(_runQueue),
+      _runQueue,
       <String, dynamic>{'timeout': 1000},
     );
   }
@@ -51,14 +48,14 @@ Future<Null> enqueueIdleTask<T>(IdleTaskHandler<T> handler, T input) {
   return task.completer.future;
 }
 
-void _runQueue(IdleDeadline deadline) {
+Future _runQueue(IdleDeadline deadline) async {
   while (!deadline.didTimeout &&
       deadline.timeRemaining() > 0 &&
       _taskQueue.isNotEmpty) {
     final task = _taskQueue.removeAt(0);
     try {
-      task.handler(task.input);
-      task.completer.complete();
+      final r = await task.handler();
+      task.completer.complete(r);
     } catch (e, t) {
       // Listeners to the completer future might throw.
       if (!task.completer.isCompleted) {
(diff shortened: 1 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/requestIdleCallbackPolyfill.js
```diff
@@ -0,0 +1,24 @@
+// A poyfill for requestIdleCallback.
+//
+// For: IE 11, Edge, and Firefox
+// Source: https://www.npmjs.com/package/requestidlecallback-polyfill
+// https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback
+window.requestIdleCallback =
```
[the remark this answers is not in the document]
toddbeckman-wf: The intention would be to add this polyfill to our `polyfill.js`, but that is private and this library is OSS so it felt wrong not to include it here. Given we already have a `polyfill.js` file to load, would there be an advantage to masking this in Dart rather than including it there?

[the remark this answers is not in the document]
maxwellpeterson-wf: We don't really have a standardized `polyfill.js` - a file with that name was added as a stopgap to address a SEV1 [here](https://github.com/Workiva/wdesk_sdk/pull/3054). Ideally this would be moved to a static asset that is loaded by the experience(s) that need it. Or maybe we would do as Rob suggested and use polyfill.io.

lib/requestIdleCallbackPolyfill.js
```diff
@@ -0,0 +1,24 @@
+// A poyfill for requestIdleCallback.
+//
+// For: IE 11, Edge, and Firefox
```
maxwellpeterson-wf: Note according to [caniuse](https://caniuse.com/#feat=requestidlecallback) Firefox supports this but Safari does not.

toddbeckman-wf: Oh, nice. I must have glanced too quickly at the table on Mozilla.

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,78 @@
...
+
+/// This calls the [requestIdleCallback] API directly.
+///
+/// Note that `requestIdleCallbackPolyfill.js` or another polyfill like it must
+/// be included in order to use this API in browsers such as IE 11 or Edge which
+/// do not support it.
+///
+/// Most tasks could be added using [enqueueIdleTask] instead.
+@JS()
+external num requestIdleCallback(Function callback, [Map options]);
```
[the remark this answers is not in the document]
toddbeckman-wf: I've changed this to use the dart bindings for it instead.

lib/requestIdleCallbackPolyfill.js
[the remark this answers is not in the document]
toddbeckman-wf: Given the use of the simpler `new Future.delayed` trick if `requestIdleFrame` isn't around, do you think we could get away with just leaving this polyfill in for OSS users but not using it in Wdesk? With Chrome and Firefox support and the new Edge announcement we may not need to add a polyfill to the code after all.

robbecker-wf:
Is there any way to cancel a queued task or raise/change the priority if needed? We're really close to writing a task scheduler here with different priority queues.

robbecker-wf:
If a cancelable completer is needed https://pub.dartlang.org/documentation/async/latest/async/CancelableCompleter-class.html
and maybe look at https://pub.dartlang.org/documentation/collection/latest/collection/HeapPriorityQueue-class.html for a priority queue

toddbeckman-wf:
@robbecker-wf I recorded your ideas in this ticket: https://jira.atl.workiva.net/browse/AF-3466

robbecker-wf:
Since standard testing may not work well, could we provide an example that allows people to manually test easier with a button to queue a new task, and a button or toggle to do an intensive task that the queued tasks would be backed up behind. Then manual testing could use dev tools to verify it works?

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,67 @@
...
+    _currentTask = window.requestIdleCallback(
+      allowInterop(_runQueue),
+      <String, dynamic>{'timeout': 1000},
+    );
+  }
+}
+
+void _runQueue(IdleDeadline deadline) {
+  while (!deadline.didTimeout &&
+      deadline.timeRemaining() > 0 &&
```
robbecker-wf: I wonder if we should give a few more ms here. Should we start another callback if we have 1ms left? Seems likely it will overflow the frame. But it obviously depends on how much work the callback is doing. 
Future: Can we send the amount of time remaining to the callback so it can bail early?

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,67 @@
...
+      _taskQueue.isNotEmpty) {
+    final task = _taskQueue.removeAt(0);
+    task.handler(task.input);
+  }
+
+  if (_taskQueue.isNotEmpty) {
+    _currentTask = window.requestIdleCallback(
+      allowInterop(_runQueue),
+      <String, dynamic>{
+        'timeout': 1000,
```
robbecker-wf: should this be configurable?

toddbeckman-wf: This doesn't need to be configurable. All this means is this is how often we will reset listening for an idle frame. It won't affect the frequency or timing of when tasks are completed.

lib/idle_task_queue.dart
toddbeckman-wf: I've added the future idea to the ticket linked above. As for your first question, a few ms might make sense here. Ideas on a range? I'd hesitate to make this more than a few ms because otherwise we might run into an issue where jobs are unexpectedly not getting done and consumers of the API would need to be made aware that the buffer exists to explain why.

robbecker-wf: Right, we don't want to starve the queue. Maybe just leave it as is for now and see if we actually ever blow out frames because of this. It could not be an issue in practice.

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,67 @@
...
+    );
+  }
+}
+
+void _runQueue(IdleDeadline deadline) {
+  while (!deadline.didTimeout &&
+      deadline.timeRemaining() > 0 &&
+      _taskQueue.isNotEmpty) {
+    final task = _taskQueue.removeAt(0);
+    task.handler(task.input);
```
robbecker-wf: What happens if the task throws an unexpected exception or error here? Will it kill the queue?

toddbeckman-wf: Good catch. That would probably break something as-is

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  _IdleTask(this.handler, this.input);
+}
+
+/// Adds the task to a queue which will be called during idle time.
+///
+/// Returns the `Future` which will complete when the task has run.
+///
+/// May complete with an error if the task throws when called.
+Future<Null> enqueueIdleTask<T>(IdleTaskHandler<T> handler, T input) {
```
greglittlefield-wf: What are your thoughts on passing through the result of the call to `handler`?

```dart
typedef FutureOr<R> IdleTaskHandler<T, R>(T input);
Future<R> enqueueIdleTask(IdleTaskHandler<T, R> hander, T input) { ... }
```

I'm thinking that might make it easier to convert existing async logic to use this utility.
```dart
Future<String> doTheThingAsync() async => 'the thing';
String doTheThing() => 'the thing';

final theThing1 = await enqueuIdleTask((_) => doTheThingAsync(), null);
final theThing2 = await enqueuIdleTask((_) => doTheThing(), null);
```

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,88 @@
...
+    _currentTask = window.requestIdleCallback(
+      allowInterop(_runQueue),
+      <String, dynamic>{'timeout': 1000},
+    );
+  }
+
+  return task.completer.future;
+}
+
+void _runQueue(IdleDeadline deadline) {
```
greglittlefield-wf: I'm not sure I'm following the purpose of this queue; doesn't [the browser do this already](x6.2-invoke-idle-callbacks-algorithm) when there are multiple idle callbacks requested?

lib/idle_task_queue.dart
greglittlefield-wf: >  It won't affect the frequency or timing of when tasks are completed.
I don't think this is true. From my understanding, this represents the longest time this task can be queued without being called.

So, if there's a lot going on and not a lot of idle periods, this will affect how often these queued tasks are completed.

I'm a little concerned about obfuscating this `timeout` as well as `timeRemaining` from the user (see also comment above about this queue).

It would be nice if the user had the option to configure this timeout, since I wouldn't expect all queued tasks to have the same priority. Perhaps we could have a sane default value that users could then configure.

lib/idle_task_queue.dart
robbecker-wf: requestIdleCallback will queue things up and not start another callback if the deadline has expired .. so yes the queue here is a little redundant. In the future, having our own queue, we could add the ability to change the order of queued items.

toddbeckman-wf:
@greglittlefield-wf I can't respond to that as a comment thread for some reason but anyway:

The purpose of the queue is to provide an API which allows tasks to be called *only* during idle time. The timeout for the queue (currently 1000ms just arbitrarily) does not cause tasks to be run; it just queues up another wait if there was no idle time. Passing the time remaining into the task handler could be an improvement to the API, but if consumers want to make use of the `timeout` they should just skip this queue and call `requestIdleCallback` directly.

greglittlefield-wf:
Right, but what advantage does the queue have vs using individual `requestIdleCallback` calls that don't set `timeout`?

lib/idle_task_queue.dart
```diff
@@ -0,0 +1,85 @@
...
+/// Adds the task to a queue which will be called during idle time.
+///
+/// Returns the `Future` which will complete when the task has run.
+///
+/// May complete with an error if the task throws when called.
+Future<R> enqueueIdleTask<R>(IdleTaskHandler<R> handler) {
+  // For unsupported browsers when the polyfill is not available,
+  // we still want the task to be called eventually.
+  if (!_doesBrowserSupportIdleCallback) {
+    return new Future.delayed(Duration.ZERO).then((_) async {
```
greglittlefield-wf: Following up the conversation here
> __Greg:__ #nit _immediately is a little misleading here, since this happens in the next tick. Also, the async modifier on this function also results in an additional async gap in Dart 1.
> 
> __Todd:__ In Dart 2 won't that cause the start of handler to be called synchronously?

No, the `new Future(callback)` constructor _"Creates a future containing the result of calling computation asynchronously with Timer.run."_ ([docs](https://api.dartlang.org/stable/2.1.0/dart-async/Future/Future.html)).

So, the function isn't run until, at the soonest, the next tick in both Dart 1 and Dart 2.

PR #105, opened 2019-02-14
Allow leaner debug mode

georgelesica-wf:
> - bug / critical
> - feature

### Description
The debug mode is very, very noisy. We'd like to be able to choose the debug "features" we want to use in a given situation.

### Changes
Add options to disable telemetry and logging (separately) in debug mode.

### Semantic Versioning

- [ ] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -161,7 +161,7 @@ class Disposable implements disposable_common.Disposable {
   /// code transpiled to JavaScript from another language.
   static void enableDebugMode({bool logging: true, bool telemetry: true}) {
     disposable_common.Disposable
-        .enableDebugMode(logging: logging, telemetry: telemetry);
+        .enableDebugMode(enableLogging: logging, enableTelemetry: telemetry);
 
     // Attach a leak flag factory function to the window to allow consumers to
     // attach leak flags to arbitrary objects.
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -160,7 +160,8 @@ class Disposable implements disposable_common.Disposable {
   /// and manually attach it to an object. For example, this may be useful in
   /// code transpiled to JavaScript from another language.
   static void enableDebugMode({bool logging: true, bool telemetry: true}) {
-    disposable_common.Disposable.enableDebugMode(logging: logging, telemetry: telemetry);
+    disposable_common.Disposable
+        .enableDebugMode(logging: logging, telemetry: telemetry);
 
     // Attach a leak flag factory function to the window to allow consumers to
     // attach leak flags to arbitrary objects.
@@ -199,23 +200,21 @@ class Disposable implements disposable_common.Disposable {
       _disposable.awaitBeforeDispose(future);
 
   @override
-  Future<Null> dispose({bool flag}) {
+  Future<Null> dispose() {
     _disposable
       ..onDisposeHandler = onDispose
       ..onWillDisposeHandler = onWillDispose;
-    return _disposable.dispose(flag: false).then((_) {
-      // We want the description to be the runtime type of this
-      // object, not the proxy disposable, so we need to reset
-      // the leak flag here.
-      if (flag != false) {
-        flagLeak(runtimeType.toString());
-      }
-    });
+    // We want the description to be the runtime type of this
+    // object, not the proxy disposable, so we need to set
+    // the leak flag here, before we delegate the `dispose`
+    // call.
+    flagLeak();
+    return _disposable.dispose();
   }
 
   @override
   void flagLeak([String description]) {
-    _disposable.flagLeak(description);
+    _disposable.flagLeak(description ?? runtimeType.toString());
   }
 
   @override
--- a/lib/src/browser/disposable_browser.dart
+++ b/lib/src/browser/disposable_browser.dart
@@ -159,8 +159,8 @@ class Disposable implements disposable_common.Disposable {
   /// generate a [LeakFlag] object. This can be used to generate a [LeakFlag]
   /// and manually attach it to an object. For example, this may be useful in
   /// code transpiled to JavaScript from another language.
-  static void enableDebugMode() {
-    disposable_common.Disposable.enableDebugMode();
+  static void enableDebugMode({bool logging: true, bool telemetry: true}) {
+    disposable_common.Disposable.enableDebugMode(logging: logging, telemetry: telemetry);
 
     // Attach a leak flag factory function to the window to allow consumers to
     // attach leak flags to arbitrary objects.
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/common/disposable.dart
```diff
@@ -58,7 +58,7 @@ class ManagedDisposer implements _Disposable {
 
   /// Dispose of the object, cleaning up to prevent memory leaks.
   @override
-  Future<Null> dispose() {
+  Future<Null> dispose({bool flag}) {
```
seanburke-wf: Is this leftovers?

lib/src/browser/disposable_browser.dart
```diff
@@ -203,17 +204,17 @@ class Disposable implements disposable_common.Disposable {
...
-    return _disposable.dispose().then((_) {
-      // We want the description to be the runtime type of this
-      // object, not the proxy disposable, so we need to reset
-      // the leak flag here.
-      flagLeak(runtimeType.toString());
-    });
+    // We want the description to be the runtime type of this
+    // object, not the proxy disposable, so we need to set
+    // the leak flag here, before we delegate the `dispose`
+    // call.
```
greglittlefield-wf: Should we call out that this may cause false positive leak flags when the object takes a while to dispose?

lib/src/browser/disposable_browser.dart
```diff
@@ -159,8 +159,9 @@ class Disposable implements disposable_common.Disposable {
   /// generate a [LeakFlag] object. This can be used to generate a [LeakFlag]
   /// and manually attach it to an object. For example, this may be useful in
   /// code transpiled to JavaScript from another language.
-  static void enableDebugMode() {
-    disposable_common.Disposable.enableDebugMode();
+  static void enableDebugMode({bool logging: true, bool telemetry: true}) {
```
seanburke-wf: `useLogging`, `useTelemetry`?

lib/src/browser/disposable_browser.dart
```diff
@@ -203,17 +204,17 @@ class Disposable implements disposable_common.Disposable {
...
+    // the leak flag here, before we delegate the `dispose`
+    // call.
+    flagLeak();
+    return _disposable.dispose();
   }
 
   @override
   void flagLeak([String description]) {
-    _disposable.flagLeak(description);
+    _disposable.flagLeak(description ?? runtimeType.toString());
```
greglittlefield-wf: IIRC `runtimeType` is very slow (and also isn't useful in minified builds). We should be extremely cautious about where we use it.

lib/src/common/disposable.dart
greglittlefield-wf: Is this not a breaking change for overriders of `dispose`?

test/unit/stubs.dart
```diff
@@ -67,7 +67,7 @@ abstract class StubDisposable implements Disposable {
 class DisposeCounter extends Disposable {
   int disposeCount = 0;
   @override
-  Future<Null> dispose() {
+  Future<Null> dispose({bool flag}) {
```
seanburke-wf: Also here.

lib/src/common/disposable.dart
```diff
@@ -630,7 +638,7 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
         _logUnmanageMessage(controller);
         _internalDisposables.remove(disposable);
       }
-      disposable.dispose();
+      disposable.dispose(flag: false);
```
seanburke-wf: Also here.

lib/src/common/disposable.dart
greglittlefield-wf: Nvm, looks like this was addressed

lib/src/browser/disposable_browser.dart
greglittlefield-wf: Oh, looks like we were already using it 😦

lib/src/browser/disposable_browser.dart
seanburke-wf: Looks like it got renamed on `Disposable` but not here.

lib/src/browser/disposable_browser.dart
georgelesica-wf: Technically that's always possible since the actual process of disposal is unrelated to whatever process actually causes the object to become available for GC, so I would argue that this doesn't really change anything for consumers.

lib/src/browser/disposable_browser.dart
```diff
@@ -159,9 +159,9 @@ class Disposable implements disposable_common.Disposable {
   /// generate a [LeakFlag] object. This can be used to generate a [LeakFlag]
   /// and manually attach it to an object. For example, this may be useful in
   /// code transpiled to JavaScript from another language.
-  static void enableDebugMode({bool logging: true, bool telemetry: true}) {
+  static void enableDebugMode({bool disableLogging, bool disableTelemetry}) {
```
seanburke-wf: It'd be good to still default these, and I prefer positive boolean names. Negating negative names gets weird. `if (enableLogging)` is easier to read than `if (!disableLogging)`.

lib/src/browser/disposable_browser.dart
georgelesica-wf: It's used in a few places in our platform pieces. Maybe we should audit that.

lib/src/browser/disposable_browser.dart
georgelesica-wf: The default is "logging is enabled", so the positive action being taken is to specifically disable it. At least that's how I think about it in a case like this. Also, default values in Dart are kind of lame because someone can still pass `null` directly (intentionally or otherwise).

seanburke-wf: This isn't an action, though, it's a bool. If it were a function `disableLogging()`, that'd be fine, but for bools I'm firmly of the belief that, without a very good reason, they should be positive in nature.

georgelesica-wf: I'm of the firm belief that this is a very good reason :-)

jeroldalbertson-wf:
+1 to these changes.

georgelesica-wf:
@Workiva/release-management-p

lib/src/browser/disposable_browser.dart
greglittlefield-wf: > the actual process of disposal is unrelated to whatever process actually causes the object to become available for GC

I would disagree with that; I thought most disposed weren't meant to be interacted with any further, and should be made available to be GC'd ASAP.

PR #107, opened 2019-03-07
Add the object's runtime type on disposal error

joshprzybyszewski-wf:
### Description
When catching `StateError`'s from disposables, it's difficult to know what class threw the error. In an effort to improve dev debugging, let's add the object's runtime type to the string.

### Changes
Add `'$runtimeType'` to the `StateError` string

### Semantic Versioning

- [x] **Patch**
- [ ] **Minor**
* [ ] Major

### Testing/QA

- [ ] CI passes

### Code Review

@dustinlessard-wf
@evanweible-wf
@jayudey-wf
@maxwellpeterson-wf
@sebastianmalysa-wf
@trentgrover-wf

### DIFF
--- a/lib/src/common/disposable.dart
+++ b/lib/src/common/disposable.dart
@@ -710,11 +710,11 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
     }
     // ignore: deprecated_member_use
     if (isDisposing) {
-      throw new StateError('$methodName not allowed, object is disposing');
+      throw new StateError('$runtimeType.$methodName not allowed, object is disposing');
     }
     if (isDisposed) {
       throw new StateError(
-          '$methodName not allowed, object is already disposed');
+          '$runtimeType.$methodName not allowed, object is already disposed');
     }
   }
 


aviary-wf:
## Security Insights

## Action Items
---

robbecker-wf:
Does this work when compiled to dart2js? When minified?
If not, it might be good to clarify that somewhere.

seanburke-wf:
We shouldn't be catching `Error`s. What's the usecase for this?

georgelesica-wf:
OK so it looks like there are two issues here. One is that `runtimeType` is slow, which honestly isn't that big a deal since it is only used when an error is thrown. The other is that `runtimeType` works different in different environments. It appears to work using DDC, but it's not clear whether it will work in Dart2JS, particularly when minified.

A better alternative here is probably to provide a field that can be overridden on classes that want to expose their type for debugging. This pattern has also been used in w_module (https://github.com/Workiva/w_module/blob/master/lib/src/lifecycle_module.dart#L251). If there's a desire to go this route I could put up a PR.

georgelesica-wf:
Here's a PR https://github.com/Workiva/w_common/pull/108

joshprzybyszewski-wf:
goal achieved in https://github.com/Workiva/w_common/pull/108

PR #124, opened 2019-06-24
CPLAT-6209 Add check for compiler to use at runtime

smaifullerton-wk:
## Motivation
  
In some cases, we need to conditionally include or exclude code and tests based on the compiler used because of known bugs with either dart2js or DDC. This PR exports a getter that can be used at runtime to determine the compiler.

## Changes
  
Use the `window.$dartLoader` global property, which is not null when DDC is used.

## Review

  

Please review: 

### QA Checklist
- [ ] CI passes
- [ ] Pull this branch, use the `isCompiledWithDdc` bool to print something in a test, run the test compiling with both DDC and dart2js and ensure that the correct thing is printed.

## Merge Checklist
- [ ] A Client Platform member has reviewed these changes


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/browser/web_compiler.dart
```diff
@@ -0,0 +1,12 @@
...
+
+import 'package:js/js.dart';
+
+@JS(r'$dartLoader')
+external dynamic get _dartLoader;
+
+// The compiler used to compile the Dart to JS, determined at runtime
+// via a global window property that is only present when DDC is used.
+// If false, DDC was used.
+bool get isCompiledWithDart2Js => _dartLoader == null;
```
evanweible-wf: Could we add a test for this? Also, this is just semantics, but it might be more accurate to expose this as a `bool get isCompiledWithDdc => _dartLoader != null;` since we're using the presence of a DDC construct, not a dart2js one.

smaifullerton-wk: Hmm. I don't know if there's really a way to test it without using it since the expectation would have to be set based on which compiler is being used... but to know which compiler is being used requires using the boolean being tested. We could assume that the test command in `tool/test.sh` is what'll always be used, but since `pub run test` uses dart2js, I don't think it's very useful to test that the property is null. Plus, if anyone used DDC the test would fail. 

It's easily manually tested, but if it ever changes, consumer CI will fail if people skip tests using it (like we'll have to in our big closed-source consumer). I'll add a more cautionary comment about using it and unfortunately I think that's the best we can do, unless you have any other ideas?

lib/src/browser/web_compiler.dart
```diff
@@ -0,0 +1,15 @@
...
+@JS(r'$dartLoader')
+external dynamic get _dartLoader;
+
+// The compiler used to compile the Dart to JS, determined at runtime
+// via a global window property that is only present when DDC is used.
+// If false, dart2js was used. Works with Dart 2 only.
+// Per the Dart team, the presence of this window property isn't
+// guaranteed forever, so proceed with caution and only use if you
+// absolutely MUST to work around compiler bugs.
+bool get isCompiledWithDdc => _dartLoader != null;
```
todbachman-wf: I'd vote for making that a `///...` doc comment. If we do that, let's make the first sentence a short summary that is a [standalone paragraph](https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph).

todbachman-wf: I think at some point we might have suggested applying `visibleForTesting` to this to try to limit use of this to unit tests. Should we do that?

lib/src/browser/web_compiler.dart
```diff
@@ -0,0 +1,16 @@
+@JS()
+library web_compiler;
+
+import 'package:js/js.dart';
+
+@JS(r'$dartLoader')
+external dynamic get _dartLoader;
+
+/// The compiler used to compile the Dart to JS, determined at runtime
+/// via a global window property that is only present when DDC is used.
```
todbachman-wf: This reads to me as though this is going to return something representing the compiler. Since it returns a boolean indicating whether it was compiled with DDC or not, maybe a sentence like, "Whether or not the Dart dev compiler (DDC), as opposed to the production compiler (dart2js), was used to compile the app." Then I think we could include the explanation that this is determined via the global window property in the paragraph below.

olesiathoms-wk:
QA +1
- [x] CI passes
- [x] Pull this branch, use the `isCompiledWithDdc` bool to print something in a test, run the test compiling with both DDC and dart2js and ensure that the correct thing is printed.

smaifullerton-wk:
@Workiva/release-management-p

PR #144, opened 2021-04-23
Migrating to Dart 2.12

alanknight-wk:
## Motivation
  
This creates a Dart 2.12 version of w_common. It is not compatible with older versions.

## Changes
  
Update the dependencies and build image, and update for breaking changes in those dependencies - specifically glob and sass.
Because this is not backward-compatible, we'd have to maintain two branches, but since this package
is rarely updated I'm considering that acceptable to do for a while.

## Review

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
- [ ] A Client Platform member has reviewed these changes


aviary2-wf:
## Security Insights

## Action Items
---

pubspec.yaml
```diff
@@ -4,28 +4,28 @@ description: General utilities for Dart projects.
 homepage: https://www.github.com/Workiva/w_common
 
 environment:
-  sdk: '>=2.4.0 <3.0.0'
+  sdk: '>=2.9.0 <3.0.0'
```
corwinsheahan-wf, line 7: Why `2.9.0` and not `2.12.0`?

pubspec.yaml
```diff
@@ -4,28 +4,28 @@ description: General utilities for Dart projects.
...
 
 dev_dependencies:
-  build_runner: ^1.7.1
-  build_test: ^0.10.9
-  build_web_compilers: ^2.5.1
-  dependency_validator: ^1.4.0
+  build_runner: '>=1.7.1 <3.0.0'
+  build_test: '>=0.10.9 <3.0.0'
+  build_web_compilers: '>=2.5.1 <4.0.0'
+  dependency_validator: '>=1.4.0 <4.0.0'
```
corwinsheahan-wf, line 29: Will this actually resolve with some of the lower versions of the version ranges here? Wondering if we can just use `^` to limit to a single major version for most/all of these.

corwinsheahan-wf (commented):
Just a couple questions on the version ranges in the pubspec.

pubspec.yaml
alanknight-wk: If we say 2.12.0 is the minimum, that opts all the code into null-safety by default, and I didn't try to do that conversion. So 2.9.0 is to prevent it being picked up from 2.7.2 users.

alanknight-wk: If it's 2.12 then it forcibly opts in all the code to null-safety. I didn't want to do that conversion as well, so 2.9 is high enough to stop 2.7.2 clients from using it, where it won't work.

pubspec.yaml
alanknight-wk: We can't do that for all of then, e.g. @#$@ sass made a breaking change in a minor version, and the build packages are weird. But I did a pub downgrade and raised the limits to what that will support.

corwinsheahan-wf (approved):

alanknight-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #198, opened 2024-11-20
FEDX-1991: allow a nullable to be managed

dustinlessard-wf:
## Motivation

When migrating disposables to null safety, it would be helpful to be able to manage a **nullable** disposable. For example:
```
 LoggingDataService? logging;
...
void _cloneBaseDataServices(AppIntelligence other) {
manageDisposable(logging);
}
```
Alternatively,
```
 LoggingDataService? logging;
...
void _cloneBaseDataServices(AppIntelligence other) {
logging = manageAndReturnTypedDisposable(other.logging?.clone(_appName));
}
```

In fact, when we inspect the `manageDisposable` [implementation](https://github.com/Workiva/w_common/blob/master/w_common/lib/src/common/disposable.dart#L557 ), we see that we have already accommodated the incoming value being null, despite our nullable-typing suggesting it cannot be.  I believe this was an oversight.

Additionally, the [comment](https://github.com/Workiva/w_common/blob/master/w_common/lib/src/common/disposable_manager.dart#L49) on `manageAndReturnTypedDisposable` suggests that it was our intention to return a null if a null was passed in as the target:
```
/// A null parameter will result in a null return value.
T manageAndReturnTypedDisposable<T extends Disposable>(T disposable);
```

## Changes

- Adjusted `manageAndReturnTypedDisposable` to allow nullables
- Added `_throwOnInvalidDisposable` which will still perform the state verification but will not throw if the value is null. (This was done to accommodate `manageAndReturnTypedDisposable`)

## Review

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
- [ ] A Frontend Architecture member has reviewed these changes

### DIFF
--- a/w_common/lib/src/browser/disposable_browser.dart
+++ b/w_common/lib/src/browser/disposable_browser.dart
@@ -243,8 +243,7 @@ class Disposable implements disposable_common.Disposable {
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
   @override
-  T manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(
-          T disposable) =>
+  T? manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(T? disposable) =>
       _disposable.manageAndReturnTypedDisposable(disposable);
 
   @override
@@ -252,7 +251,7 @@ class Disposable implements disposable_common.Disposable {
       _disposable.manageCompleter(completer);
 
   @override
-  void manageDisposable(disposable_common.Disposable disposable) =>
+  void manageDisposable(disposable_common.Disposable? disposable) =>
       _disposable.manageDisposable(disposable);
 
   @override
--- a/w_common/lib/src/browser/disposable_browser.dart
+++ b/w_common/lib/src/browser/disposable_browser.dart
@@ -243,7 +243,8 @@ class Disposable implements disposable_common.Disposable {
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
   @override
-  T? manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(T? disposable) =>
+  T? manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(
+          T? disposable) =>
       _disposable.manageAndReturnTypedDisposable(disposable);
 
   @override
--- a/w_common/lib/src/browser/disposable_browser.dart
+++ b/w_common/lib/src/browser/disposable_browser.dart
@@ -243,9 +243,10 @@ class Disposable implements disposable_common.Disposable {
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
   @override
-  T? manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(
-          T? disposable) =>
-      _disposable.manageAndReturnTypedDisposable(disposable);
+  T manageAndReturnTypedDisposable<T extends disposable_common.Disposable?>(
+      T disposable) {
+    return _disposable.manageAndReturnTypedDisposable<T>(disposable);
+  }
 
   @override
   Completer<T> manageCompleter<T>(Completer<T> completer) =>
--- a/w_common/lib/src/common/disposable.dart
+++ b/w_common/lib/src/common/disposable.dart
@@ -515,7 +515,8 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
   @mustCallSuper
   @override
   T? manageAndReturnTypedDisposable<T extends Disposable>(T? disposable) {
-    _throwOnInvalidDisposable('manageAndReturnDisposable', 'disposable', disposable);
+    _throwOnInvalidDisposable(
+        'manageAndReturnDisposable', 'disposable', disposable);
     manageDisposable(disposable);
 
     return disposable;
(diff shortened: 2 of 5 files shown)


dustinlessard-wf:
Consumer analysis: https://github.com/Workiva/gha-dart/actions/runs/11938183028

dustinlessard-wf:
1 valid failure found ... https://github.com/Workiva/gha-dart/actions/runs/11938183028/job/33276016036

w_common/lib/src/common/disposable.dart
```diff
@@ -514,8 +514,9 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
 
   @mustCallSuper
   @override
-  T manageAndReturnTypedDisposable<T extends Disposable>(T disposable) {
-    _throwOnInvalidCall('manageAndReturnDisposable', 'disposable', disposable);
+  T manageAndReturnTypedDisposable<T extends Disposable?>(T disposable) {
```
todbachman-wf, line 517: **Suggestion**: Here too I think it might read better this way (though it just might be the way my brain works):

```suggestion
  T? manageAndReturnTypedDisposable<T extends Disposable>(T? disposable) {
```

w_common/lib/src/common/disposable.dart
```diff
@@ -552,12 +553,11 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
...
   }
 
-  void manageDisposable(Disposable disposable) {
-    // ignore: unnecessary_null_comparison
+  void manageDisposable(Disposable? disposable) {
     if (disposable == null) {
       return;
     }
-    _throwOnInvalidCall('manageDisposable', 'disposable', disposable);
+    _throwOnInvalidDisposable('manageDisposable', 'disposable', disposable);
```
todbachman-wf, line 560: **Question**: Is there a reason you need to add this new function? I took a look at the existing one and it seems like it would still work for us in this situation.

w_common/lib/src/common/disposable_manager.dart
```diff
@@ -47,7 +47,7 @@ abstract class DisposableManagerV7 {
   ///      }
   ///
   /// A null parameter will result in a null return value.
-  T manageAndReturnTypedDisposable<T extends Disposable>(T disposable);
+  T manageAndReturnTypedDisposable<T extends Disposable?>(T disposable);
```
todbachman-wf, line 50: **Suggestion**: 😸 

```suggestion
  T? manageAndReturnTypedDisposable<T extends Disposable>(T? disposable);
```

w_common/lib/src/common/disposable.dart
dustinlessard-wf: The existing one does a null check on the incoming value and is used for non-Disposable values as well.

todbachman-wf: We're doing a null check right above this line so we won't call that function if disposable is null. So it looks like we could still use the existing function. Were you seeing test errors or something using the existing one?

w_common/lib/src/browser/disposable_browser.dart
```diff
@@ -243,16 +243,17 @@ class Disposable implements disposable_common.Disposable {
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
   @override
-  T manageAndReturnTypedDisposable<T extends disposable_common.Disposable>(
-          T disposable) =>
-      _disposable.manageAndReturnTypedDisposable(disposable);
+  T manageAndReturnTypedDisposable<T extends disposable_common.Disposable?>(
+      T disposable) {
```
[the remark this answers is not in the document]
dustinlessard-wf: This is what I had originally, but in a group code review with FEDX, we landed on the current form because we didn't want to require that consumers do this:
```
var foo = manageAndReturnTypedDisposable(Foo());
foo?.bar();
```
We want consumers to be able to do:
```
var foo = manageAndReturnTypedDisposable(Foo());
foo.bar();
```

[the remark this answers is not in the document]
todbachman-wf: Oh interesting! So they are not equivalent. Nice, I like the ergonomics of that better. Let's keep it as you have it.

w_common/lib/src/common/disposable.dart
```diff
@@ -514,7 +514,11 @@ class Disposable implements _Disposable, DisposableManagerV7, LeakFlagger {
 
   @mustCallSuper
   @override
-  T manageAndReturnTypedDisposable<T extends Disposable>(T disposable) {
+  T manageAndReturnTypedDisposable<T extends Disposable?>(T disposable) {
+    if (disposable == null) {
+      return disposable;
+    }
+
     _throwOnInvalidCall('manageAndReturnDisposable', 'disposable', disposable);
```
todbachman-wf, line 522: It looks like we have a typo here. Maybe we should fix it while we're in here.

```suggestion
    _throwOnInvalidCall('manageAndReturnTypedDisposable', 'disposable', disposable);
```

w_common/lib/src/common/disposable.dart
dustinlessard-wf: Fair - I didn't have a 'good' reason for adding a new function and agree, adding a null check while preserving the original function usage is totally fine.

todbachman-wf (approved):

todbachman-wf:
QA +1.

dustinlessard-wf:
Robots wait!

todbachman-wf (approved):

robbecker-wf:
QA+1 CI passes and tests are sufficient

rmconsole-wf (approved):
+1 from RM

